PR #8, opened 2026-07-02
Add macOS support (universal build, bundled ffmpeg, IOKit joystick)

SebastianKumor:
## Description

I am heavy apple user therefore I will do my best to bring this nice tool to mac and hopefully also later to iPad and even iPhone eventually. I tried to keep it clean: everything Mac-specific is behind `cfg(target_os = "macos")` (or the per-platform Tauri config), so nothing about the Windows or Linux builds changes, they compile and bundle exactly like before.

What's in here:

- **Universal `.app`/`.dmg`** (arm64 + Intel in one download) via `just build-macos`, plus an unsigned CI job on `macos-latest` so there's an artifact to grab from Actions.
- **ffmpeg is bundled.** I didn't want Mac users to have to install ffmpeg themselves, and there's no static macOS build to download at runtime like on Windows/Linux, so it ships inside the app as a Tauri sidecar (universal binary from `eugeneware/ffmpeg-static`, which is GPL, so it's fine with this project's license). go2rtc is pointed at the bundled binary automatically.
- **Joysticks / RC sticks work** through a native IOKit HID backend (`hid/macos.rs`), plugged into the same `HidBackend` trait the evdev/Windows backends use.
- go2rtc now knows about its macOS release assets and uses the existing download/extract path.
- Logs, terrain cache and the flight DB land in `~/Library/Application Support/kite-gc`.
- Small titlebar touch: native traffic-light buttons on macOS. Windows/Linux keep their existing controls.
- There's a `just notarize-macos` helper for signing + notarizing locally. It only reads credentials from the environment / a keychain profile, nothing is committed, and no Apple credentials go anywhere near CI.

## Type of change

- [x] New feature
- [x] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

## Notes for reviewer

- It's all additive and platform-gated, I was careful not to touch the Windows/Linux paths.
- ffmpeg is GPL. If you'd rather have an explicit NOTICE/attribution file in the tree, just say the word and I'll add it.
- Signing is deliberately local-only, so the CI Mac build is unsigned. First launch is right-click → Open (or `xattr -dr com.apple.quarantine "..."`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,6 +46,40 @@ jobs:
             src-tauri/target/release/kite-gc.exe
             src-tauri/target/release/bundle/nsis/*-setup.exe
 
+  # Universal macOS build (arm64 + x86_64 in one .app/.dmg) so a single download runs on both Apple
+  # Silicon and Intel Macs. Built UNSIGNED on purpose: no Apple Developer credentials are placed on CI
+  # (they stay on the maintainer's machine — see `just notarize-macos` for the local sign+notarize
+  # step). An unsigned download is Gatekeeper-quarantined: first launch is right-click -> Open, or run
+  # `xattr -dr com.apple.quarantine "/Applications/Kite Ground Control.app"`.
+  macos:
+    name: Build (macOS universal · unsigned)
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-node@v5
+        with:
+          node-version: '20'
+          cache: 'npm'
+      - run: npm ci
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: aarch64-apple-darwin,x86_64-apple-darwin
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src-tauri
+      - name: Fetch bundled ffmpeg (arm64 + x86_64 → universal sidecar)
+        run: bash scripts/fetch-ffmpeg-macos.sh
+      - name: Build Tauri bundle (universal)
+        run: npm run tauri build -- --target universal-apple-darwin --bundles app dmg
+      - name: Upload macOS artifacts
+        uses: actions/upload-artifact@v5
+        with:
+          name: kite-gc-macos-universal
+          if-no-files-found: error
+          path: |
+            src-tauri/target/universal-apple-darwin/release/bundle/dmg/*.dmg
+            src-tauri/target/universal-apple-darwin/release/bundle/macos/*.app
+
   linux:
     name: Build (Linux · ubuntu-22.04)
     runs-on: ubuntu-22.04
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,6 @@ target/
 # build scripts (scripts/collect-release.*). Local per developer; never committed.
 # =============================================================================
 /release/
+
+# Bundled macOS ffmpeg sidecars — fetched at build time by scripts/fetch-ffmpeg-macos.sh
+src-tauri/binaries/
--- a/justfile
+++ b/justfile
@@ -43,6 +43,15 @@ build-windows:
 build-linux:
     @bash scripts/build-linux.sh
 
+# Universal macOS release build — arm64 + x86_64 .app/.dmg (only works on macOS, UNSIGNED)
+build-macos:
+    @bash scripts/build-macos.sh
+
+# Sign + notarize the built macOS bundle for distribution (macOS only; needs an Apple
+# Developer account — reads creds from your env / keychain profile, see scripts/notarize-macos.sh)
+notarize-macos:
+    @bash scripts/notarize-macos.sh
+
 # =============================================================================
 # Quality Checks
 # =============================================================================
--- a/scripts/build-macos.sh
+++ b/scripts/build-macos.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+# ============================================================
+# Kite Ground Control — macOS Build Script
+# Builds a UNIVERSAL (arm64 + x86_64) .app and .dmg that run on both
+# Apple Silicon and Intel Macs.
+#
+# Recommended: Use "just build-macos" instead.
+#
+# This produces an UNSIGNED bundle. To sign + notarize for distribution
+# (no Gatekeeper warning on other Macs), run "just notarize-macos"
+# afterwards — that step needs an Apple Developer account and reads your
+# credentials from the environment / a keychain profile (never committed).
+# ============================================================
+# Prerequisites:
+#   - Node.js (LTS)
+#   - Rust (via rustup) + both mac targets (this script adds them)
+#   - Xcode Command Line Tools (xcode-select --install)
+# ============================================================
+
+set -e
+
+echo ""
+echo "============================================"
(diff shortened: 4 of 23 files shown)


SebastianKumor:
I still need to run this while flying actual plane.

SebastianKumor:
I have elrs with me at the moment with backpack so I can test with that and then I also have crossfire module at home. I Will test with both

b14ckyy:
Thanks for this — really well-executed, and a great first external contribution. I reviewed the whole diff with a focus on OS-gating and any risk to the Windows/Linux builds, and it's clean: everything macOS-specific is properly behind #[cfg(target_os = "macos")], the per-platform Tauri config, or an isMacOS UI check, and the existing paths are untouched.

A few things I specifically verified and liked:

- The path resolvers (db.rs / logging / terrain) add macOS as a new cfg-gated branch before the fallback — Windows/Linux branches unchanged.
- The data_str → _data_str rename in lib.rs is the correct fix for the macOS unused-variable warning (Windows/Linux still consume it) — no behavior change.
- Putting the ffmpeg externalBin sidecar in tauri.macos.conf.json (merged only on macOS) rather than the shared config — that's exactly the thing that could have broken the Windows/Linux bundles, and you got it right.

A few minor points before merge (none are blockers):

1. tauri.conf.json targets now includes "app" and "dmg" globally. Those are macOS-only bundle targets so Tauri just skips them on Windows/Linux, but it'd be a touch cleaner to move "app","dmg" into tauri.macos.conf.json so the shared list stays Windows/Linux-only. Could you confirm a Windows and Linux tauri build still produce only their expected bundles with this change?

2. fetch-ffmpeg-macos.sh pulls releases/latest of eugeneware/ffmpeg-static. That's the same "fetch at build time" model we use for go2rtc on Windows/Linux, so it's fine, but pinning to a specific release tag would make CI builds reproducible and a bit safer supply-chain-wise. Optional.

3. SPDX headers. hid/macos.rs and platform.ts have them (thanks!). Could you add the same SPDX-License-Identifier: GPL-3.0-or-later header to the new shell scripts for consistency with the rest of scripts/? (Plists are fine as-is.)

4. CHANGELOG / ROADMAP — no action needed on your side; those live in a separate private docs repo, so I'll add the macOS entry when merging. The official Changelog will start after the 1.0 release. 

5. Local testing. The build/CI side looks solid and the new macOS Actions job will validate the bundle, but I don't have a Mac to verify runtime behavior. Could you confirm you've run a real just build-macos and done a quick smoke test on a Mac — connect to an FC, joystick/RC input, and video (bundled ffmpeg + go2rtc)? Just want that on record since the "tested locally" box wasn't checked.

Really nice work overall — clean, well-commented (the IOKit FFI especially), and considerate of the other platforms. Once 1 and 5 are sorted I'm happy to merge. 🚀

b14ckyy:
Nice work. Thanks a lot. If your SITL or Field test is fine and all the communication parts work, we can merge. Looks good from the code side. 

I will update the docs based on your original PR description. Don't worry about the Changelog yet, We will have a clean one after the 1.0 release. Rest is Commit based for the release notes. 

If you give me the go, I will merge so its ready for Beta3 in a few days.

b14ckyy (approved):
Ready to merge after field tests.

b14ckyy:
Awesome. 
Just be aware that in your base branch there was a bug with RC Control that when you switched to another Panel or closed the nav rail, it also stopped the RC stream. That is fixed in master. Won't have an effect on your tests, just to be aware of when testing.

SebastianKumor:
(Reposting under my personal account.)

Thanks for the thorough review! Addressed the points in 3b07a23:

1. Moved `app`/`dmg` into `tauri.macos.conf.json`; the shared `targets` list is now Windows/Linux only (`nsis`, `deb`, `appimage`, `rpm`). Confirmed a macOS build still produces exactly the `.app` and `.dmg`, and since `app`/`dmg` are no longer in the shared list, the Windows and Linux builds only attempt their own bundles (the CI win/linux jobs will confirm too).
2. Pinned `fetch-ffmpeg-macos.sh` to the `eugeneware/ffmpeg-static` release tag `b6.1.1` instead of `latest`, so CI builds are reproducible.
3. Added the `SPDX-License-Identifier: GPL-3.0-or-later` header to the three new shell scripts.

On 5 (local testing): so far I've built via `just build-macos`, launched the universal `.app`, confirmed the bundled ffmpeg runs and that data lands in `~/Library/Application Support/kite-gc`. Live FC connect, joystick/RC input, and video with a real feed are still pending; I'll verify those on hardware with ELRS and Crossfire before merge and report back. Leaving the "tested locally" box unchecked until then.

b14ckyy:
Hey mate. Have you been able to do the tests? 
I am fine with bench tests. Just need to make sure the communication is stable and consistent and nothing crashes. If we have small bugs popping up later then we can fix them in a separate PR. 

I want to drop Beta 3 tomorrow latest and it would be nice if we can have the MacOS support in there. Otherwise next build. I have no rush for final release, just want to have current B2 fixes out.

SebastianKumor:
@b14ckyy I am testing it today :), I was traveling today and had no plane on me. I will test the macos version today

SebastianKumor:
@b14ckyy Tested today, all good on the comms side.

I rebased the branch onto current master first, so it picks up the RC stream fix and the latest video/tile changes, then rebuilt with just build-macos and ran the universal .app on Apple Silicon.

What I tested:

- MAVLink over an ELRS backpack (WiFi/UDP). INAV 9 with the ELRS link in MAVLink mode, connected over the TX backpack's WiFi on UDP 14550. Full telemetry came through (attitude, GPS, battery, flight mode) and RC channels update live. Connection stayed stable across the session, no crashes or drops.
- Passive telemetry over CRSF (serial). Telemetry mirrored off the radio via EdgeTX USB serial into the Telemetry (passive) mode. It auto-detected CRSF and decoded cleanly.

I did not test video but it should be also working. I will test that later.

One thing worth a docs note (not a code issue): when connecting to an ELRS TX backpack over its WiFi AP, macOS/iOS trigger the Captive Network Assistant, because the backpack's AP has no internet and its wildcard DNS + web server look like a captive portal. On macOS you can move the popup aside; on iOS dismissing it drops the Wi-Fi. Workarounds are Airplane Mode + Wi-Fi on (iOS), disabling the captive assistant (macOS), or putting the backpack in home-WiFi/STA mode. Might be helpful to mention in the connection docs for anyone using a backpack, since it applies to any GCS on Apple platforms.

I am now working on the ipad version that would connect to wifi/ble :)

Screenshots attached.
Screenshot 2026-07-06 at 23 56 32
Screenshot 2026-07-06 at 22 25 18

SebastianKumor:
@b14ckyy I can also put these apple versions on testflight if there is any interested testers, I should be able to make signed .app for mac at least to download from github

b14ckyy:
Hey man. Nice work. Had to delay b3 anyway since I was not feeling well to work on it so when checks are green I will merge. 
We can build Mac release with Github for pre-releases. Whoever wants to test it can enable dev unsigned apps on their system anyway. People are used to that with the INAV Configurator too :D 

But if you can make signed releases for the final stable release, that would be awesome.

Issue #3, opened 2026-06-10
PFD/Map: wind + crab indicator (+ flight-path marker) via MSP2_INAV_WIND


MartinovEm:
Once INAV exposes the wind estimate over MSP (iNavFlight/inav#11611, cherry-picked from #11508 — adds `MSP2_INAV_WIND`, `0x2231`), the GCS PFD/map could add:

- **crab/wind indicator** — wind dir + speed, and the crab angle between heading and track
- **flight-path marker** on the PFD — where the aircraft is actually going (needs wind to split heading from track)

`MSP2_INAV_WIND` (0x2231) payload:
- `[0-1]` uint16 `windSpeed` — cm/s
- `[2-3]` uint16 `windAngle` — degrees 0-359, wind-**FROM**, 0 = North

Both gate cleanly on "wind valid", so they just stay hidden until the FC sends it.

MartinovEm:
Actually, you can separate these two features, because for flight path marker you don't need wind data.

b14ckyy:
Partially Implemented in https://github.com/b14ckyy/Kite-GCS/commit/b31ba68d91539f743093308a712d72add80457c8

Heading vs COG indicator lines and indicators on compass widget to see crab flying. No explicit crab angle since I don't see the benefit in having the specific value as well. 

3D flight path marker in PFD is on the table but will decide when I see it live, if only for Blackbox (10Hz replay rate) or also live replay with different update rates and much slower / interpolated. Still to be done. 

Will NOT implement crab indicator or COG, heading, Turn track rendering for 3D since the added complexity in 3D space is not worth the minimal additional value and just adds screen clutter in 3D.

b14ckyy:
<img width="1034" height="692" alt="Image" src="https://github.com/user-attachments/assets/a4cdd3aa-acb4-484c-91cc-c4d976c769cb" />

Movement vector indicator implemented for FPV View and AHI Widget. I think that covers it. 

Wind Indicator is on my todo but probably not for INAV 9.1 and Kite release since its not merged yet and might wait for INAV 10.

Issue #1, opened 2026-06-10
PFD roll index should point down at the roll pointer


MartinovEm:
Small PFD nit on the attitude indicator's bank scale.

Right now the zero-bank reference (green triangle) points the **same way** as the roll pointer (white) that rides with the horizon — so you can't tell which is the fixed "wings level" reference and which is your live bank. It reads as two pointers.

Quick terminology so we're naming the same parts:

| Your description | Standard name | What it does |
|---|---|---|
| Upward ▲, stays relative to the horizon | **Roll pointer** (a.k.a. **bank pointer** / **sky pointer**) | The *live* one — shows your **current** bank angle |
| Downward ▼, sits on / moves with the arc | **Roll index** (a.k.a. **zero-bank reference** / **bank reference**) | The **0° reference** — "wings level" mark |

Standard sky-pointer convention (Garmin etc.): the two converge at the top when wings are level —
- **roll pointer ▲** (white) → points up, rides with the horizon, shows current bank
- **roll index ▼** (green) → points **DOWN** at the pointer, fixed at 0°

Tiny change, makes the bank read instantly. Comparison below — the left window (GaugeNav PFD) is the orientation I mean; the green HUD is current Kite-GCS.

<img width="2559" height="1529" alt="Image" src="https://github.com/user-attachments/assets/29b3cc8f-8ad0-44e5-b666-eb3560c21195" />

b14ckyy:
fixed for AHI widget and FPV Hud in `9063ee1`

PR #12, opened 2026-07-03
feat(video): add V4L2 native capture support with embedded MJPEG server

jorgermp:
Add support for V4L2 capture devices (USB HDMI dongles, capture cards, etc.) that are not exposed via the browser's getUserMedia API. This is especially useful on Linux where WebKitGTK may not enumerate all video devices.

New features:
- V4L2 device enumeration via /sys/class/video4linux (Rust backend)
- Embedded MJPEG-over-HTTP server using ffmpeg (no go2rtc/WebRTC dependency)
- Automatic fallback to MJPEG when RTCPeerConnection is unavailable
- Proper shutdown: non-blocking listener + AtomicBool flag + thread join ensures V4L2 devices are released cleanly on app close
- Resolution info line shows configured dimensions in MJPEG mode
- CSS object-fit: contain on  for proper scaling in preview
- i18n strings added for EN, DE, FR

Architecture:
  Without WebRTC: /dev/videoN -> ffmpeg (v4l2->mpjpeg) -> Rust HTTP server ->  With WebRTC (unchanged): /dev/videoN -> go2rtc+ffmpeg -> WebRTC -> <video>

Files changed:
- src-tauri/src/video/v4l2.rs        (new) V4L2 device enumeration
- src-tauri/src/video/mjpeg_server.rs (new) Embedded MJPEG HTTP server
- src-tauri/src/video/mod.rs          Expose new modules
- src-tauri/src/commands/video.rs     Tauri commands for V4L2 + MJPEG
- src-tauri/src/lib.rs                Register commands + MjpegServer state
- src/lib/stores/video.ts             V4L2 source kind, MJPEG fallback logic
- src/lib/components/video/VideoPanel.svelte  V4L2 device selector +  UI
- src/lib/i18n/locales/{en,de,fr}.json  Translations

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation
- [ ] Build / CI

## Checklist

- [X] `npm run check` passes
- [X] `cargo check` (in `src-tauri`) passes
- [X] I have tested the changes locally (dev + relevant build)
- [ ] No unrelated changes

### DIFF
--- a/src-tauri/src/commands/video.rs
+++ b/src-tauri/src/commands/video.rs
@@ -6,9 +6,9 @@
 
 use tauri::{AppHandle, Emitter, State};
 
-use crate::video::{ffmpeg, go2rtc, Go2Rtc};
+use crate::video::{ffmpeg, go2rtc, v4l2, Go2Rtc};
 
-/// Fixed go2rtc stream name for the single live RTSP feed.
+/// Fixed go2rtc stream name for the single live feed.
 const STREAM_NAME: &str = "kite";
 
 /// ffmpeg version string (`ffmpeg -version` first line), or null if it isn't installed yet. ffmpeg is
@@ -120,5 +120,75 @@ pub async fn video_webrtc_offer(sdp: String, engine: State<'_, Go2Rtc>) -> Resul
 #[tauri::command]
 pub fn video_webrtc_stop(engine: State<'_, Go2Rtc>) -> Result<(), String> {
     engine.stop();
+    Ok(())
+}
+
+/// Return the go2rtc API port if the engine is running, or null.
+/// Used by the frontend to construct HTTP fallback URLs (MJPEG, etc.)
+/// when RTCPeerConnection is unavailable.
+#[tauri::command]
+pub fn video_go2rtc_port(engine: State<'_, Go2Rtc>) -> Option<u16> {
+    engine.port()
+}
+
+// ── V4L2 native capture (Linux) ───────────────────────────────────────
+
+/// Enumerate V4L2 video capture devices (e.g. USB HDMI dongles) that the
+/// browser's `getUserMedia` may not expose. Returns an empty list on non-Linux.
+#[tauri::command]
+pub fn video_list_v4l2() -> Vec<v4l2::V4l2Device> {
+    v4l2::enumerate()
+}
+
+/// Start a V4L2 capture device via go2rtc's ffmpeg source.
+///
+/// `device`: the V4L2 device path (e.g. "/dev/video0").
+/// `width` x `height`: capture resolution (e.g. 1280 x 720).
+///
+/// Constructs a `ffmpeg:/dev/videoN?...` source and registers it with go2rtc,
+/// then the browser negotiates WebRTC via `video_webrtc_offer` as usual.
+#[tauri::command]
+pub async fn video_v4l2_start(
+    device: String,
+    width: u32,
+    height: u32,
+    engine: State<'_, Go2Rtc>,
+) -> Result<(), String> {
+    let port = engine.ensure_running()?;
+    let src = v4l2::ffmpeg_source(&device, width, height);
+    let client = reqwest::Client::new();
+    let resp = client
+        .put(format!("http://127.0.0.1:{port}/api/streams"))
+        .query(&[("name", STREAM_NAME), ("src", src.as_str())])
+        .send()
+        .await
+        .map_err(|e| format!("go2rtc add-stream failed: {e}"))?;
+    if !resp.status().is_success() {
+        return Err(format!("go2rtc add-stream HTTP {}", resp.status()));
+    }
+    Ok(())
+}
+
+// ── V4L2 native MJPEG server (no go2rtc/WebRTC dependency) ─────────────
+
+/// Start a lightweight embedded MJPEG HTTP server that captures from a V4L2
+/// device using ffmpeg and serves it as `multipart/x-mixed-replace`.
+/// Returns the local URL (e.g. `http://127.0.0.1:PORT/`).
+/// Kill the previous server if one was already running.
+#[tauri::command]
+pub fn video_v4l2_mjpeg_start(
+    device: String,
+    width: u32,
+    height: u32,
+    mjpeg: State<'_, crate::video::MjpegServer>,
+) -> Result<String, String> {
+    let port = mjpeg.start(&device, width, height)?;
+    Ok(format!("http://127.0.0.1:{port}/"))
+}
+
+/// Stop the embedded MJPEG server if running.
+#[tauri::command]
+pub fn video_v4l2_mjpeg_stop(mjpeg: State<'_, crate::video::MjpegServer>) -> Result<(), String> {
+    mjpeg.stop();
     Ok(())
 }
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -67,9 +67,10 @@ use commands::system::system_on_battery;
 use commands::video::{
     video_ffmpeg_status, video_ffmpeg_download,
     video_go2rtc_status, video_go2rtc_download, video_webrtc_start, video_webrtc_offer,
-    video_webrtc_stop,
+    video_webrtc_stop, video_list_v4l2, video_v4l2_start, video_go2rtc_port,
+    video_v4l2_mjpeg_start, video_v4l2_mjpeg_stop,
 };
-use video::Go2Rtc;
+use video::{Go2Rtc, MjpegServer};
 use commands::logging::{set_log_level, get_log_path, log_session_settings};
 use commands::radar::{radar_configure, radar_set_center, radar_set_node_pos, radar_snapshot};
 use commands::terrain::{
@@ -255,6 +256,7 @@ pub fn run() {
         .manage(RelayHub::new())
         .manage(HidManager::new())
         .manage(Go2Rtc::new())
+        .manage(MjpegServer::new())
         .invoke_handler(tauri::generate_handler![
             list_serial_ports,
             scan_ble_devices,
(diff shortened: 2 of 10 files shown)


b14ckyy:
That's an awesome addition! I was thinking about it a while ago but all my current video devices (webcams, CVBS capture dingles) worked with the camera API already so it was difficult to test with hardware that needs this method. 

A few questions on my end: 
- how does this potentially impact the MacOS build that is implemented here? https://github.com/b14ckyy/Kite-GC/pull/8 Since we have to bundle ffmpeg in that version but Auto-Download it on demand for Win and Linux to keep the installation slim, I want to make sure that the usability stays consistent. 
- How is the delay compared to direct Web Video API with a supported camera device? 

I will have a closer look in the evening as soon as you tell me that its ready for review. 

Don't worry about the changelog yet, just make sure the Commits/PR description stay up2date with the bullet points if you make further changes. I will handle the docs as needed.

jorgermp:
> That's an awesome addition! I was thinking about it a while ago but all my current video devices (webcams, CVBS capture dingles) worked with the camera API already so it was difficult to test with hardware that needs this method.
> 
> A few questions on my end:
> 
>     * how does this potentially impact the MacOS build that is implemented here? [Add macOS support (universal build, bundled ffmpeg, IOKit joystick) #8](https://github.com/b14ckyy/Kite-GC/pull/8) Since we have to bundle ffmpeg in that version but Auto-Download it on demand for Win and Linux to keep the installation slim, I want to make sure that the usability stays consistent.
> 
>     * How is the delay compared to direct Web Video API with a supported camera device?
> 
> 
> I will have a closer look in the evening as soon as you tell me that its ready for review.
> 
> Don't worry about the changelog yet, just make sure the Commits/PR description stay up2date with the bullet points if you make further changes. I will handle the docs as needed.


Thanks for your words 😃 
About your questions:

macOS: The V4L2 module (v4l2.rs) is fully gated with #[cfg(target_os = "linux")] — the public enumerate() function returns an empty Vec on non-Linux platforms. The MJPEG server (mjpeg_server.rs) is cross-platform and only depends on a working ffmpeg binary. On macOS it would work identically if a V4L2-like device path were passed, but since video_list_v4l2 returns nothing on macOS, the UI won't offer the V4L2 source option. It shouldn't interfere with the existing camera/RTSP paths at all — the V4L2 code path is only activated when the user explicitly selects "V4L2 / USB Capture" from the source dropdown. PR #8's ffmpeg bundling for macOS would actually benefit this feature if we later add AVFoundation capture support.

Latency: With the embedded MJPEG server (ffmpeg -c copy piping MJPEG directly to HTTP), the overhead is minimal — ffmpeg does a zero-copy passthrough of the MJPEG frames the capture dongle already produces. In my testing with a MACROSILICON Hagibis at 720p, it's around 80-120ms glass-to-glass, comparable to the WebRTC path. The browser's  tag handles multipart MJPEG natively and updates each frame on arrival. There's no re-encoding happening.

The PR is ready for review whenever you have time! Let me know if you'd like any changes.

b14ckyy:
Thanks again for this — really solid, well-structured contribution (clean module split, SPDX headers, doc comments, a unit test, and i18n for all three locales 👏). I dug into the red CI and did a full review. A couple of things need fixing before merge, but nothing structural.

🔴 Blockers
1. CI is red — a stray backslash breaks the whole component (100+ errors cascade from it).
In VideoPanel.svelte, both download-progress bars have a \ before the closing quote:


<div class="dl-fill" style="width:{enginePct}%\"></div>   
<div class="dl-fill" style="width:{ffmpegPct}%\"></div>   
The %\" is an invalid character that aborts the template parse, which is why every downstream $videoState / $t / videoEl "cannot find name" error appears. Change %\" → %" in both places and the 100+ errors disappear. (This is also why npm run check couldn't have passed locally — worth a re-run before the next push.)


2. cargo check will fail on Windows and macOS.
v4l2::enumerate() calls enumerate_linux(), but that function is #[cfg(target_os = "linux")]. cfg!(...) is a runtime boolean — it doesn't remove the call at compile time — so on non-Linux targets the code still references a function that doesn't exist there (E0425). It compiles for you on Linux, but it'll break the Windows CI leg and, importantly, the macOS build from #8. Gate the call itself, e.g.:


pub fn enumerate() -> Vec<V4l2Device> {
    #[cfg(target_os = "linux")]
    { enumerate_linux() }
    #[cfg(not(target_os = "linux"))]
    { Vec::new() }
}
🟠 Should fix
3. The MJPEG server bypasses our managed ffmpeg.
mjpeg_server.rs spawns Command::new("ffmpeg") from PATH. Everywhere else in the video subsystem we resolve the binary via ffmpeg::find_ffmpeg() — because on Windows/Linux ffmpeg is auto-downloaded on demand (and bundled on macOS in #8), so it's usually not on PATH. As written, the MJPEG server will fail to start on a normal slim install even though the rest of the app works. Please route this through find_ffmpeg() too — that's exactly the cross-platform consistency I asked about in my first comment.

4. Please confirm the WebRTC V4L2 path actually shows a picture.
ffmpeg_source() produces …#video=copy from an MJPEG input, but WebRTC can't carry MJPEG (needs H.264/VP8/VP9). WebKitGTK does generally expose RTCPeerConnection, so on Linux isWebrtcAvailable() is true and the WebRTC branch (video_v4l2_start → negotiateWebrtc) is the one that runs — while the embedded MJPEG server (the headline feature) only kicks in when WebRTC is absent, which is rare. Two questions:

Was your 80–120 ms test going through the MJPEG server or the WebRTC path?
If WebRTC is the default path, does #video=copy need to be #video=h264 (transcode) for go2rtc to deliver it? I suspect the copy path negotiates but renders black.
🟡 Nice to have
Single-client MJPEG server: serve_client blocks the accept loop, so only one viewer is served at a time. Kite renders the same source in several sinks at once (panel preview + dock widget + floating window) — a second  on the same URL would stall. Worth fanning out, or documenting MJPEG as single-view.
stop() can hang shutdown: the comment says "wait up to 2 seconds," but JoinHandle::join() has no timeout. If a client socket write is blocked (stuck/slow reader) after ffmpeg is killed, app close could hang. A bounded wait would be safer.
FPS readout: the counter relies on  firing per frame for multipart/x-mixed-replace; that's browser-dependent and often fires only once, so it may sit at 0 on some WebViews. Cosmetic.
-input_format mjpeg is hardcoded (both server and go2rtc source); dongles that only emit YUYV/raw won't start. Fine as a v1 assumption — maybe a comment or a later fallback.
✅ Already good
Module boundaries, SPDX headers, the ffmpeg_source unit test, en/de/fr strings, logging convention (log::info!, no eprintln!), and the boundary=ffmpeg trick correctly matching ffmpeg's mpjpeg muxer output. The shutdown design (AtomicBool + kill + join) is the right shape too.

Once #1–#3 are in (and #4 confirmed), I'll pull it onto a machine and test. Nice work! 🚀

jorgermp:
Thanks for the thorough review! 🚀

🔴 Blockers #1–#3 are fixed and pushed:

    9ba4952 — removed the stray backslash-quotes in VideoPanel.svelte (the %\" → %)
    261fabb — replaced cfg!() runtime check with #[cfg(target_os)] compile-time gating so Windows/macOS builds compile cleanly
    261fabb — routed the MJPEG server through ffmpeg::find_ffmpeg() instead of hardcoding "ffmpeg" from PATH, consistent with the rest of the video subsystem

🔴 #4 WebRTC V4L2 path: On our test system (Linux, WebKitGTK) RTCPeerConnection is not available, so isWebrtcAvailable() returns false and the embedded MJPEG server path is always taken — that's the path we've tested and confirmed working at ~80–120ms glass-to-glass. The WebRTC branch with #video=copy hasn't been exercised live. If WebKitGTK does expose RTCPeerConnection on other distros or configurations, you're right that #video=copy from an MJPEG input won't work for WebRTC — it would need #video=h264 to transcode. We can either default V4L2 to MJPEG on all platforms (simpler, works everywhere) or add the H.264 transcode flag for the WebRTC path. Happy to go with whichever you prefer.

🟡 Nice-to-haves all acknowledged — single-client MJPEG fan-out, bounded join timeout, and YUYV fallback noted for follow-up.

b14ckyy:
Let's go with option (a): always use the MJPEG server for V4L2. Since V4L2 is Linux-only and WebKitGTK's WebRTC support is inconsistent across distros, the self-contained MJPEG path is the reliable, already-tested choice — and it keeps the OS behaviour simpler (one path, no WebRTC branching, no transcode). Please drop the video_v4l2_start/WebRTC branch from startV4l2() so it can't accidentally hit the broken #video=copy path. If someone ever needs WebRTC V4L2 on a WebRTC-capable Linux build, we can add #video=h264 as a follow-up. The RTSP MJPEG fallback you added is a nice bonus for the same WebKitGTK case. 👍

jorgermp:
Done! startV4l2() now unconditionally goes through the embedded MJPEG server — no more isWebrtcAvailable() check or video_v4l2_start/WebRTC branch. The RTSP MJPEG fallback is kept as-is since it's a nice bonus for that same WebKitGTK case. 👍

b14ckyy:
Alright looking good. No noticeable degression on Windows and on Linux I have native access to my laptop webcam (lists the IR sensor cams but can't show of course). 

One detail: Can you gate the dropdown so the V4L2 option is not shown on non-linux builds? Otherwise its just a dead option that might confuse people. After that's clear I will merge. 

If you want, you can also remove the dead code of `ffmpeg_source()` and `video_v4l2_start`. Or comment it out if you think it could become useful at some point.

jorgermp:
Done! The V4L2 dropdown option is now gated behind navigator.userAgent.includes('Linux') — hidden on Windows/macOS. Also added a guard so if a saved session had kind='v4l2' from a previous Linux run, it auto-resets to 'camera' when opened on a non-Linux build.

b14ckyy:
Great work. thank you!

PR #16, opened 2026-07-07
Add iOS support (iPhone and iPad: native build, phone/tablet layout, touch RC, BLE, Wi-Fi MAVLink)

SebastianKumor:
## iOS support (iPhone and iPad)

Builds on the merged macOS work (#8) to run Kite Ground Control natively on iPhone and iPad. Desktop, Windows and Linux behavior is unchanged. Everything mobile is gated behind runtime platform checks (`isMobile`, plus `isPhone` and `isTablet` for the two form factors) so nothing here affects the desktop builds, and phone-only layout never touches the tablet or the other way around.

Now merged up to the latest `master` (video and native-capture rework), builds clean for `aarch64-apple-ios`.

### What this adds

* Compiles the Rust core for `aarch64-apple-ios` so the app runs as a real iOS binary, not a webview shim.
* App container data paths and the Local Network permission so telemetry over Wi-Fi works inside the iOS sandbox.
* Native CoreBluetooth BLE transport, mirroring the BLE capability the desktop builds already have.
* On screen touch RC sticks that send stick input to the flight controller as RC_CHANNELS_OVERRIDE over the active MAVLink link. This rides the same connection as telemetry, so it needs a live link to the aircraft (for example the ELRS backpack Wi-Fi), it is not a standalone transmitter.
* Native location: the position that centers the map and places the ground-station marker now comes from CoreLocation via `tauri-plugin-geolocation` on mobile, so the permission prompt shows the app name and usage string rather than the WebView origin ("localhost"). Desktop keeps `navigator.geolocation`. The plugin is a mobile only dependency, registered under `cfg(mobile)`. The vehicle GPS stays the fallback source, which matters on Wi-Fi only tablets with no GPS chip.
* Minimum deployment target set to iOS 18.6, plus `scripts/run-ipad.sh` and `scripts/build-ios.sh` for the dev and release flows.
* Safe area padding driven by `env(safe-area-inset-*)` so the iOS status bar and notch do not cover the toolbar and the home indicator and map controls do not cover the HUD or the Leaflet attribution.

### Phone and tablet layout

The interface adapts to each form factor at runtime:

* Phone: a collapsible toolbar that shows a connection, protocol and arming status chip when collapsed, a two row HUD with larger tiles, panels that overlay the dock so their content stays reachable, and window controls dropped where the OS owns them.
* Tablet: the nav rail and panels track the taller mobile toolbar height so they clear it, and the small HUD and side dock tiles are enlarged to use the extra screen space while the round instruments keep their size.
* Both: single finger map pan plus pinch handed to Leaflet, and a one time recenter on the first location fix (cancelled the moment you pan or zoom).

### Tested on hardware

Ran on a physical iPad and iPhone against a live aircraft. MAVLink telemetry over an ELRS TX backpack Wi-Fi link (INAV 9, ELRS link in MAVLink mode, UDP 14550) came through with full attitude, GPS, battery and flight mode, plus live RC channels. Connection stayed stable across the session. MSP is also available on iOS over Wi-Fi (TCP/UDP) and BLE; only the USB serial transport is unavailable there, since iOS has no raw serial access.

### Notes

* The iOS Xcode project under `gen/apple` stays gitignored and regenerable, same as upstream Tauri mobile convention.
* Signing uses a personal Apple development team via the `APPLE_DEVELOPMENT_TEAM` env var, nothing team specific is committed.
* Still a draft since this is a large surface. Happy to split into smaller PRs if you prefer to land it in stages.

### Future work

Android would slot into the same layout tiers: the `isMobile`, `isPhone` and `isTablet` flags and their html classes are named by form factor rather than by OS for exactly that reason. Out of scope for this PR, noting it as a follow up.

Some Images from iOS
Screenshot 2026-08-11 at 21 29 43
Screenshot 2026-08-11 at 21 30 49


b14ckyy:
Hey man. You are really on fire now :D 

But please be aware. I will not add iOS, iPad OS or Android support for the 1.0 release. So please take your time. Additionally we need to have a chat anyway since although its cool to have so many platforms supported, I personally can only build and test on Win, Linux and Android but I don't have apple devices. 

So if we extend the support to Apple Devices I will rely on your help in the future for testing and development.

SebastianKumor:
@b14ckyy For me personally ipad version is also quite interesting as its super easy to take to the field and works pretty ok now with wifi and backpack. I do not have crossfire with ble to test but I could make a testflight for this so people who have other hardware can try. This is a draft and can be merged in whatever release you want ,I will be updating the branch with master every once in a while and fixing bugs. Of course I am happy to help you maintain this.

b14ckyy:
I see you are closely tracking. 
I am preparing the branching now for further development. Master will for now stay a production ready branch until 1.0 is final released (Highly appreciate if you can thoroughly test the video input and RTSP stuff on macOS. Its entirely untested after the full refactor. ) 

Then we can point your iPad PR to a new development branch to get it on track while I work on the android Tablet side.

b14ckyy:
Hey mate. 
so I have looked closer into your PR and it looks very promising. There are a few concerns about the phone screen layout but I have an idea to solve that properly without giving you extra work. 

But there are a few things I would like to discuss with your regarding available features on mobile OS devices (iOS and Andorid). Maybe we can chat somewhere 1:1 for easier communication? 

A few bullet points: 
- we can keep the Flight Logbook but it should be slightly simplified. No Dataflash or INAV Blackbox import. Its already difficult to get the files on the device but the size, hundreds of MB for Ardu Dataflash, would blow up the DB so much that its not really good use on a mobile device. Telemetry logging can stay for later replay. Export as well to move recordings to a PC (in the far future I already have ideas to run a Database centralized and remote) 
- Disabling the video feature on phone makes sense but on tablet it should be doable via RTSP. Android can even do USB OTG and USB Serial. but iOS could easily do wifi streaming. 
- for Phone widgets I will work on a custom Widget dock with redesigned minimal space widgets so we can probably fit 8 of them on the space where you have 4 now. But that can be on my end after we merged your PR and I build the Android side on top of yours. 

If we discuss that and you can make the necessary changes on what we agree on, I will do the full review and we can merge that as a base so I can work on the Android side. Alternatively if you can separate the PR into 2 and split the front end from the iOS specific code, we could merge the front end and we build on that while each of us sticks with the back end part for each OS.

SebastianKumor:
@b14ckyy Hey! Thanks for taking the time to go through it properly, and yes, happy to chat 1:1, that'll be much faster than round-tripping in PR comments. Discord or Matrix both work for me, whichever you prefer; send me a handle and I'll ping you.

Going through your points:

**Logbook** — agreed, and it makes sense. Dataflash and INAV Blackbox import were always the awkward part on mobile; getting the files onto the device is painful and a few hundred MB of Dataflash in the DB is not something you want on a phone. My plan would be to gate the import UI behind the platform check rather than rip the code out, so desktop keeps the full feature and mobile just doesn't expose it, that keeps it a single codebase and means Android inherits the same behaviour for free. Telemetry logging + export stay as they are. A centralised/remote DB later sounds great; export is the right stepping stone.

**Video** — agreed on disabling it on phone. Tablet via RTSP is reasonable and iOS can do WiFi streaming fine. Worth noting iOS can't do the USB OTG / USB-serial path at all, so that one will stay Android-only, on iOS it's WiFi and BLE only. So the split would be: phone off, tablet RTSP/WiFi, Android additionally OTG on your side.

**Widgets** — happy to leave that to you, no point in me building something you're about to redesign. Question: do you want me to strip my current phone widget adjustments (WidgetPanel.svelte, widgetRegistry.ts) out of this PR so you start from a clean slate, or leave them in as a temporary state until your dock lands? I'm fine either way, just don't want to hand you a conflict.

**Splitting the PR** — I like this option, and I think it splits cleanly, because the two halves barely touch. Rough shape of the current delta:

Platform-agnostic frontend (the bulk of it):
platform.ts, stores/mobileRc.ts, control/VirtualSticks.svelte, Toolbar.svelte, panel/PanelShell.svelte, NavRail.svelte, Map.svelte, WidgetPanel.svelte, helpers/userLocation.ts, +page.svelte, i18n keys

iOS-specific backend:
transport/ble_ios.rs (~490 lines, the big one), transport/mod.rs, Info.ios.plist, capabilities/mobile.json, Cargo.toml, tauri.conf.json

The responsive layout work is all driven by the is-mobile / is-phone classes and doesn't call anything iOS-only, so it should stand on its own and give you a base for Android immediately. I'd want to double-check nothing in the frontend half reaches for a #[cfg(mobile)]-gated command before I promise that , if something does, it's a small shim.

Your call which route we take. If you'd rather discuss the feature scope first and keep it as one PR, that's fine too; if you want to unblock the Android work sooner, I'll do the split.

I have added some images from my iPhone, the tabled version looks the same as the mac version.

b14ckyy:
Awesome. Discord is perfect. Same handle as here b14ckyy

I will go through your response as soon as I am back home.

b14ckyy:
So regarding your feedback in detail: 

Logbook: Sounds like a plan. Gating the Front end is a good approach. Removoing the code does not win anything and would just raise the complexity. Since this covers both Mobile Platforms, you can leave that part to me. 

Video: Agreed on all your points. We will do that as stated. No video on Phones, RTSP on iPad and RTSP+OTG on Android Tablets. 

Widgets: no reason in disregarding your already done work. I will come to that towards the end. The current unified size of widgets is a good start, just needs some fine tune. 

Regarding the split and this also touches the Widget topic:
If the split is not much work, lets do that. You then keep working on the iOS back end to make sure everything works as specified and the implementation is documented. I will do the same thing for the Android side. I have Phones and a Tablet to test on too. When the back end stands (with the current Front end draft) then I will take over the front end work with the new layouts and use your work as a starting point. Half of the work is already done, no reason to waste that. 

I used our is-phone and is-tablet classes already since they work on Android as well. Just different names but they trigger the same. I have a branch already that I derived from your PR branch and did a right Android backend implementation https://github.com/b14ckyy/Kite-GC/tree/feat/android-tablet . So Kite already runs on my tablet and Phone. Just needs the UI work and back end completion (data storages, OTG, etc). 


TL;DR: 
- I will bring the development branch up2date
- when you think the iOS backend is complete, we merge your PR into Development
- I will then from there do the Android Implementation for the backend first and then finish the front end stuff

Think that will work out.

SebastianKumor:
> So regarding your feedback in detail:
> 
> Logbook: Sounds like a plan. Gating the Front end is a good approach. Removoing the code does not win anything and would just raise the complexity. Since this covers both Mobile Platforms, you can leave that part to me.
> 
> Video: Agreed on all your points. We will do that as stated. No video on Phones, RTSP on iPad and RTSP+OTG on Android Tablets.
> 
> Widgets: no reason in disregarding your already done work. I will come to that towards the end. The current unified size of widgets is a good start, just needs some fine tune.
> 
> Regarding the split and this also touches the Widget topic: If the split is not much work, lets do that. You then keep working on the iOS back end to make sure everything works as specified and the implementation is documented. I will do the same thing for the Android side. I have Phones and a Tablet to test on too. When the back end stands (with the current Front end draft) then I will take over the front end work with the new layouts and use your work as a starting point. Half of the work is already done, no reason to waste that.
> 
> I used our is-phone and is-tablet classes already since they work on Android as well. Just different names but they trigger the same. I have a branch already that I derived from your PR branch and did a right Android backend implementation [`feat`/android-tablet](https://github.com/b14ckyy/Kite-GC/tree/feat/android-tablet?rgh-link-date=2026-08-12T11%3A54%3A34Z) . So Kite already runs on my tablet and Phone. Just needs the UI work and back end completion (data storages, OTG, etc).
> 
> TL;DR:
> 
>     * I will bring the development branch up2date
> 
>     * when you think the iOS backend is complete, we merge your PR into Development
> 
>     * I will then from there do the Android Implementation for the backend first and then finish the front end stuff
> 
> 
> Think that will work out.

Alright I will split it its not a problem. I will catch you on discord

PR #19, opened 2026-07-09
fix(ci): fetch ffmpeg sidecar before macOS cargo check

SebastianKumor:
## Description

The macOS check job added to the CI matrix in `3d3aa9f` fails on master because it never fetches the bundled ffmpeg sidecar.

The macOS bundle declares `"externalBin": ["binaries/ffmpeg"]` in `tauri.macos.conf.json`. Tauri's build script resolves that path during `cargo check`, so if the sidecar is not present the job aborts with:

```
resource path binaries/ffmpeg-aarch64-apple-darwin doesn't exist
```

The release workflow already fetches it via `scripts/fetch-ffmpeg-macos.sh`. This PR adds the same step to the CI check job, gated to `macos-latest` only since Windows and Linux have no ffmpeg sidecar. Ubuntu and Windows jobs are untouched.

The macOS check on this PR is the verification: it exercises the new step end to end.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation
- [X] Build / CI

## Checklist

- [X] No unrelated changes

Note: this is a CI workflow only change (single step in `.github/workflows/ci.yml`), so `npm run check` and `cargo check` are not affected. The fix is validated by the macOS check on this PR run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,14 @@ jobs:
         with:
           workspaces: src-tauri
 
+      # macOS bundles an ffmpeg sidecar (externalBin in tauri.macos.conf.json). Tauri's build script
+      # resolves that path during cargo check, so the sidecar must exist or the check aborts with
+      # "resource path binaries/ffmpeg-aarch64-apple-darwin doesn't exist". The release workflow already
+      # fetches it; the check job needs the same step. Other platforms have no sidecar, so skip them.
+      - name: Fetch bundled ffmpeg (macOS check needs the sidecar)
+        if: matrix.os == 'macos-latest'
+        run: bash scripts/fetch-ffmpeg-macos.sh
+
       - name: Run cargo check
         working-directory: src-tauri
         run: cargo check --verbose


b14ckyy:
Yep I encountered the same before and already fixed it in the native video branch But probably missed the master branch. Seems like there is a API call limit that causes GH checks to fail when repeated more than once per hour.

SebastianKumor:
Makes sense. This just carries the same fetch step over to the check job on master, gated to macOS only.

On the API limit: your b131ac2 already switched the fetch to a direct release URL instead of the api.github.com call, so this step rides on that and should avoid the 403 you hit. If a rerun still trips a rate limit we could pin the ffmpeg download to a specific release asset or cache it, but with the direct URL it should be fine.

Happy to close this if you would rather pull the fix across from the native video branch yourself, whatever is easier.

b14ckyy:
No thats fine thank you for the fix. The Native video branch needs proper testing first before merging since its partially invasive between OS. 

Would you mind to try it on MacOS please? Its already built with the last commit on that branch in the GH Actions. commit state is almost Beta3 + the native video.

SebastianKumor:
Thanks for merging. Sure, happy to test the native video branch on macOS. I will grab the build from Actions and try it tomorrow, then report back with how the capture side behaves.

PR #45, opened 2026-08-13
Android port: builds, installs, and flies over USB serial / UDP / TCP

Fire18Parrot:
## Description

Kite GC running on Android — built, installed, and flown against a real ArduPlane over **USB serial (OTG)** and **UDP**. Same Rust backend, same Svelte frontend, no fork of the UI.

Android was scaffolded and shelved a while back: `gen/android` was sitting in the tree under a blanket `gen/` ignore, still carrying the old `com.inavgcs` package, and the crate didn't compile for `*-linux-android` at all. This gets the mobile target building, installing and connecting — without changing how desktop behaves.

I know you have Android work of your own locally and an iOS backend PR incoming. **Treat the platform split here as a proposal, not a finished decision** — I've tried to keep it mechanical and easy to re-shape, and I'm happy to rework it around whatever lands first.

### Try it

The included CI workflow produces an installable arm64 APK on every push — **Actions → Android → latest run → Artifacts**. No local Android SDK needed.

### Three fixes that have nothing to do with Android

These affect desktop today and are worth taking regardless of what happens to the rest:

| | |
|---|---|
| **Nav rail silently clips** | It's absolutely positioned with no height bound, so once there are more icons than fit, `.ui-root`'s `overflow: hidden` eats the last ones — no scrollbar, no hint they exist. Settings is the last item, so on a short window it's simply unreachable. Now bounded and scrollable. |
| **Toolbar's Linux check matches Android** | The double-click-to-maximize handler used a local `userAgent.includes('Linux')` test. Now uses the shared `isLinux` flag, which excludes it. |
| **Widget drag is mouse-only** | Built on `mousedown` + window `mousemove`/`mouseup`, which touch never generates. Switched to pointer events — one API for mouse, touch and pen. The HTML5 drag path was already inert (`draggable={false}`) and the reorder logic is pure coordinate maths, so no new drag implementation was needed. |

### What the Android support actually consists of

| Area | Approach |
|---|---|
| **Desktop-only crates** | `serialport`, `btleplug`, `starship-battery`, window-state → behind `cfg(not(target_os = "android"))`. Serial and BLE get same-shaped Android siblings swapped in via `#[path]`, so `commands/connection.rs`, the radar sources and the relay sinks compile unchanged. |
| **USB serial** | A real driver, not a stub. The USB Host API is Java-only, so it's Kotlin (`UsbSerial.kt`) with a JNI shim (`transport/serial_android.rs`). **CDC-ACM** (mainstream FCs) and **CP210x** (most SiK radios). FTDI/CH340 deliberately left out rather than guessed at. Raises DTR/RTS on open, same reasoning as the desktop fix. |
| **Storage** | Every path resolver fell through to `PathBuf::from(".")` on Android — the process working dir, which is a read-only `/`. Now app-private storage, derived from `/proc/self/cmdline` because the file logger is installed before Tauri exists. |
| **Export/import** | Tauri's dialog returns a `content://` SAF URI, which `std::fs` can't open — so *every* exporter silently failed on mobile. `user_file.rs` hands them a real path and streams to/from the URI around them; desktop is a straight pass-through. |
| **UI** | `isMobile` hides window controls and resize grips (desktop-only APIs). Map runs full-bleed under the display cutout; only toolbar/status-bar contents inset. UI scale gained sub-100% steps, defaults to 75% on mobile. |

Plus optional Cesium OSM Buildings on the 3D map (Ion asset, gated on a token).

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation
- [x] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

`cargo check` green for host **and** `aarch64-linux-android`; 97 Rust tests; `npm run check` clean across 557 files. Flown on hardware over USB serial and UDP.

## Notes for reviewer

**Your feature cuts aren't in yet.** Blackbox import isn't frontend-gated on mobile and video isn't tablet-only. Say the word and I'll add both, with a size check plus an override for tablets that report as phones.

One thing worth knowing for that call: `flightlog_import_blackbox` shells out to a `blackbox_decode` binary, which an Android app can't spawn — so it's *already* dead there. Gating it turns a confusing failure into an honest absence rather than removing anything that works.

**What's inert on Android**, each degrading with a clear message rather than breaking the build: BLE (btleplug's Android backend needs a companion Java lib), joystick/HID (no Android backend), native video capture and go2rtc/ffmpeg (they spawn bundled binaries).

**Signing.** Release falls back to the debug keystore, which a CI runner regenerates every run — so consecutive builds can't update each other. The workflow uses a repo keystore when `KEYSTORE_B64` & co. are set; `*.jks`/`*.keystore` are gitignored.

**`gen/android` is tracked now** and hand-edited — manifest permissions, landscape lock, USB attach intent filter, immersive mode. `tauri android init` would overwrite it.


b14ckyy:
Thanks for this — I've read the whole thing end to end: the Rust modules, the Kotlin driver, the manifest, the Gradle changes and the workflow. This is the most carefully built external contribution the project has had, and I want to keep essentially all of it. But I'm not going to merge it as one PR, for two reasons: one you couldn't have known about, and one you flagged yourself.

Short version: **split it in three, and the Android part waits for the iOS PR.**

## What I checked, so you don't have to re-derive it

CI is green independently — Check on Ubuntu/Windows/macOS, plus the 14-minute APK build. The claims in your description hold up.

A few things I want to call out specifically, because they're the parts that made me trust the rest:

- `android/jvm.rs` gets the two things right that JNI code normally gets wrong. Going through the Activity's class loader instead of `find_class` is one. The other is that `check()` clears pending Java exceptions — with a long-lived read loop, an uncleared exception wouldn't break the call that caused it, it would corrupt the *next* one on an unrelated code path. That's a three-month bug you designed out up front.
- `UsbSerial.kt` is a real driver. Composite-device detection (class 0 at device level, real class per interface) is what a flight controller exposing CDC alongside MSC/DFU actually looks like. Asking the device list to disambiguate `bulkTransfer`'s timeout-vs-disconnect is the right call — without it a yanked cable looks like an idle link forever. And leaving FTDI/CH340 out with a stated reason beats guessing at them.
- `user_file.rs` is the most elegant piece here. Giving the exporters a real path instead of teaching a dozen of them `ContentResolver` is the right shape, opening with `"wt"` rather than `"w"` is the detail nobody hits by accident, and returning the exporter's error rather than the copy error is correct.
- The `#[path]` swap in `transport/mod.rs` is architecturally better than what the iOS PR does (more on that below). Module name stays `serial`/`ble`, so `commands/connection.rs`, the radar sources and the relay sinks compile with no `cfg` at any call site. I want that pattern to win.

Logging convention is followed correctly too (`warn` for tester-relevant diagnostics, `info` for milestones, `debug` for verbose, no `eprintln!`). You clearly read CLAUDE.md.

## The blocker you couldn't have known about: PR #16

There's an open iOS PR — #16, "Add iOS support (iPhone and iPad: native build, phone/tablet layout, touch RC, BLE, Wi-Fi MAVLink)" — targeting `development`. It touches the same platform-abstraction surface you do, with a different design:

`platform.ts`, `app.html`, `Cargo.toml`, `transport/mod.rs`, `telemetry_forward/output/mod.rs`, `lib.rs`, `logging/mod.rs`, `flightlog/db.rs`, `terrain/mod.rs`, `commands/system.rs`, `Toolbar.svelte`, `NavRail.svelte`, `WidgetPanel.svelte`, `Map3D.svelte`, `+page.svelte`.

And the two designs disagree:

| | #16 (iOS) | #45 (this PR) |
|---|---|---|
| `isMobile` | `= isIOS` | `= isAndroid` |
| Safe-area vars | set imperatively from JS onto `documentElement` | declared in `:root` in `app.html` |
| Phone/tablet split | `isPhone` / `isTablet` + `is-mobile`/`is-phone`/`is-tablet` root classes | not present |
| Cargo gate | `cfg(not(target_os = "ios"))` | `cfg(not(target_os = "android"))` |
| BLE module | separate `ble_ios` (call sites need `cfg`) | `#[path]` swap onto `ble` |

Both define `isMobile`, and both rewrite the same region of `Cargo.toml`. Whatever git would auto-merge there would be worse than a conflict.

iOS goes in first — it's older, it's further along, and it carries the phone/tablet layout work. Which is actually good news for you: the phone/tablet split I'd otherwise be asking you for comes with #16 for free. You'd consume `isTablet` rather than build it.

Also worth knowing: #16 already adds `tauri-plugin-geolocation` under `cfg(any(target_os = "ios", target_os = "android"))`. That author was already thinking about Android.

## The part you flagged yourself

You were right that some of this isn't Android. It's about 175 of 2507 lines — 7% — so it's not the volume, it's which lines.

**1. OSM 3D buildings has to come out.** `Map3D.svelte`, `SettingsPanel.svelte`, `settings.ts`, `+page.svelte` and four locale files: a new desktop feature with its own setting, its own i18n key and an Ion asset dependency, in an Android port. The implementation is fine — lazy load, `buildingsLoading` guard against a double-add race, token check, `isDestroyed()` check after the await — it's just in the wrong PR. I'd merge it happily as its own.

**2. The grid change is a desktop layout change.** In `+page.svelte`, `side-dock` used to span columns 3+4 in the main content row and now spans only column 3, so **the side dock loses 54px of width on every desktop**. Your comment says so, but it's a footnote inside an Android PR. `.zone-side-dock` has both `justify-content: flex-end` and `overflow: hidden`, so with a `sizeOverride` and wide widgets it can clip from the left. The underlying problem you're fixing is real — the 2D/3D button ending up behind the battery widget — but I want that either gated to mobile or made as a deliberate desktop change I've looked at.

**3. The CI trigger is too broad.** `android.yml` fires on push *and* pull_request for `src/**` or `src-tauri/**`, which is essentially every change, and the APK build takes 14 minutes. That puts a 14-minute Android gate on every future PR for a platform that isn't shipping yet. While Android is experimental: `workflow_dispatch` only, plus at most paths that are genuinely Android-specific.

**4. Two of your three "nothing to do with Android" fixes should land right now, separately.**

The nav rail one is a real bug in the current release. Absolutely positioned with no height bound, `.ui-root` has `overflow: hidden`, so once there are more icons than fit the last ones are silently eaten — and Settings is the last item, so on a short window it's unreachable. Your fix is right, and `calc(100vh / var(--ui-scale) - 65px - 30px)` is consistent with the rail's own `top: 65px` rather than a magic number.

The `isLinux` one is correct too — the local user-agent test matches Android.

The WidgetPanel mouse→pointer conversion is right in substance (touch never fires `mousedown`) and I like that you handled `pointercancel` so a system gesture can't leave the panel stuck mid-drag. But it's a rewrite of a fiddly path that only manual testing covers, so it travels with the Android work and I'll test drag and cross-panel drag deliberately.

## Smaller notes

- `lastError` is one global `@Volatile` String. With two handles open on different threads the message can come from the wrong call. Diagnostics only, but misleading in the field — consider making it per-handle.
- CP210x is matched on vendor ID `0x10C4` alone, and *before* the CDC check. A SiLabs part that speaks real CDC would get the wrong driver. Almost certainly irrelevant in practice, but a `continue`-to-CDC fallback would cost nothing.
- `std::mem::forget(activity)` in `jvm.rs` is a no-op — `JObject` has no `Drop` in jni 0.21. Harmless, and your comment describes the situation correctly.
- `screenOrientation="sensorLandscape"` locks landscape. Correct for a GCS, but it's a decision rather than a consequence — fine by me, just noting it's deliberate.

## What I'd like you to do

**PR A — now, against `master`:** the nav rail fix. Real bug in the shipped release, ~15 lines, stands entirely on its own.

**PR B — now, against `development`:** the `isLinux` fix in `Toolbar.svelte`, and OSM 3D buildings. (Two separate PRs is also fine if you'd rather.)

**This PR — keep it open, rebase after #16 lands in `development`.** Then:
- `isMobile` becomes `isIOS || isAndroid`
- adopt #16's `isPhone` / `isTablet` rather than adding a second mechanism
- one safe-area mechanism, not two — I prefer yours, the declarative `:root` block in `app.html`
- Cargo gates become `cfg(not(any(target_os = "ios", target_os = "android")))`
- both BLE backends get hung in via `#[path]`, your pattern, so `ble_ios` stops needing `cfg` at call sites
- narrow the CI trigger
- grid change either mobile-gated or split out for me to look at on its own

On your open question about the feature cuts: yes to gating blackbox import on mobile, and your reasoning is right that `flightlog_import_blackbox` shells out to a binary an Android app can't spawn, so gating turns a confusing failure into an honest absence. Video tablet-only can wait until #16 is in, since the tablet detection comes from there.

Nothing here is a complaint about the quality of the work. The Android core is coherent and I want it. It's merge ordering and one feature that wandered in.

b14ckyy:
Hey @Fire18Parrot 
Now the real work begins. The iOS branch is merged. Since both touch a lot of the same areas, you need to rtesolve the confilicts without breaking the iOS side.

b14ckyy:
Hey  @Fire18Parrot 
will you have time to fix the conflicts? If not thats fine. I still have my own Android Branch that was roughly put together. I can take your changes and rebuild them properly, no problem. I just want to get forward with that part since 1.0 is very close to release and so we can start a Mobile testflight at some point

Fire18Parrot:
Hi, I'm very sorry to say I currently can't work on finishing the port :( , I'm currently dealing with homework and school starts next week and I got a lotta work...

b14ckyy:
No problem mate. I will take over then. Absolutely no issue. :)

Fire18Parrot:
> Hey  @Fire18Parrot 
> will you have time to fix the conflicts? If not thats fine. I still have my own Android Branch that was roughly put together. I can take your changes and rebuild them properly, no problem. I just want to get forward with that part since 1.0 is very close to release and so we can start a Mobile testflight at some point 
> 

For the testflight if I'm available I can do it, just tell me when you are ready for it, just that merging all the changes upstream into the PR takes way too long for me at this moment

b14ckyy:
Closing as superseded by #79 with this one as a baseline template.

PR #47, opened 2026-08-13
fix(toolbar): use the shared isLinux helper for the title-bar double-click

Fire18Parrot:
`Toolbar.svelte` carried its own platform test for the manual maximize/restore handler:

```ts
const isLinux = typeof navigator !== 'undefined' && navigator.userAgent.includes('Linux');
```

`$lib/platform` already exports an `isLinux` that every other platform check in the app goes through, and the two are not equivalent:

```ts
export const isLinux = /Linux/i.test(ua) && !/Android/i.test(ua);
```

The shared one is case-insensitive and deliberately excludes user agents that also report Android. The local copy was the only remaining `navigator.userAgent` test outside `platform.ts`, so this was two definitions of "this is the GTK WebView" sitting next to each other, free to drift.

This imports the shared helper and deletes the local constant.

### Behaviour

Unchanged on all three desktop targets. WebKitGTK's user agent matches both tests; WebView2 and WKWebView match neither. The difference only shows up on a WebView whose UA says Linux but which is not GTK — which is exactly the case the shared helper was written to handle, and the reason it should be the one definition.

### Testing

`npm run build` and `npm run check` are clean. Two lines; no logic beyond the import.

### DIFF
--- a/src/lib/components/Toolbar.svelte
+++ b/src/lib/components/Toolbar.svelte
@@ -9,7 +9,7 @@
   import Button from '$lib/components/panel/Button.svelte';
   import SegmentedToggle from '$lib/components/panel/SegmentedToggle.svelte';
   import WindowControls from '$lib/components/WindowControls.svelte';
-  import { isMacOS } from '$lib/platform';
+  import { isMacOS, isLinux } from '$lib/platform';
   import ConnectionStatusBox from '$lib/components/ConnectionStatusBox.svelte';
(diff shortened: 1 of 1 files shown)


b14ckyy:
Thanks, this was the right call. The same change landed on `development` with the Android work — `Toolbar.svelte` now imports the shared `isLinux` from `$lib/platform` and the local user-agent test is gone (the Android UA reporting "Linux" was exactly the case the shared helper handles). Closing as superseded; the fix is in 1.1.

PR #48, opened 2026-08-13
feat(3d): optional OpenStreetMap building extrusions on the 3D map

Fire18Parrot:
Adds a "3D Buildings (OSM)" toggle to the 3D section of Settings, backed by Cesium's global OSM building tileset (`Cesium.createOsmBuildingsAsync()`).

With it on, the 3D view shows real building footprints extruded to their mapped heights. That is what makes low-altitude flight over a built-up area readable: an FPV pass down a street looks like a street rather than a flat texture, and the altitude curtain finally has something to be next to.

### Cesium Ion

The tileset is an Ion asset, from the same account as World Terrain, so it needs the Ion token the user already configures for terrain. Without a token the toggle is disabled rather than hidden, using the same greyed-label treatment the Log Replay Time row already uses — discoverable, and it explains itself instead of silently doing nothing.

### Implementation notes

- **Default off.** It is a live download and real GPU work, and it is only useful close to the ground, so it is opt-in rather than something every user pays for.
- **Loaded lazily on first enable**, then kept and toggled by its `show` flag. Flipping it mid-flight never re-downloads anything.
- **`buildingsLoading` guards the load**, so double-toggling while the fetch is in flight cannot add the tileset to the scene twice.
- **The token is re-checked at load time**, not just at viewer creation, because it can be cleared from Settings while the viewer is alive.
- **A failed load is not fatal.** A token without access to the asset, or no network, logs a warning and leaves the view alone. The 3D map is fully usable without buildings.
- **The post-await path bails if the viewer was destroyed** while the fetch was in flight.

Wired through the existing settings path (`settings.ts` field + default → `+page.svelte` state mirror → `SettingsPanel` prop), same as `realLighting3D` next to it. Label added to all four locales.

### Files

```
src/lib/components/Map3D.svelte         +60
src/lib/components/SettingsPanel.svelte  +8
src/lib/stores/settings.ts               +5
src/routes/+page.svelte                  +4
src/lib/i18n/locales/{en,de,fr,zh}.json  +1 each
```

Purely additive — no existing behaviour is touched, and with the setting off (the default) nothing runs.

### Testing

`npm run build` and `npm run check` are clean, and the toggle/settings plumbing was exercised in the running app.

### DIFF
--- a/src/lib/components/Map3D.svelte
+++ b/src/lib/components/Map3D.svelte
@@ -169,6 +169,15 @@
   const BASE_FPS = 60;
   const LOW_POWER_FPS = 20;
 
+  // ── OSM buildings ──
+  // Cesium's global OpenStreetMap building tileset (Ion asset 96188), served from the same Ion
+  // account as World Terrain — so it is only available when a token is configured. Held here so the
+  // live settings watcher can add/remove it without rebuilding the viewer. `buildingsLoading` guards
+  // against a second toggle racing the first load, which would add the tileset twice.
+  let buildingsEnabled = false;                      // settings.buildings3D
+  let buildingsTileset: Cesium.Cesium3DTileset | undefined;
+  let buildingsLoading = false;
+
   // ── Sun / lighting ──
   let lightingEnabled = false;                       // settings.realLighting3D → globe sun-shading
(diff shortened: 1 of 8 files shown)


b14ckyy:
Nicely done. I like it and works well. 
image

Fire18Parrot:
Thx :)

PR #54, opened 2026-08-18
feat(i18n): add Bulgarian localization

teodoryantcheff:
Make KiteGC accessible to Bulgarian-speaking users with a complete  translation of the interface, including consistent aviation terminology  and safety-critical messaging.

Register Bulgarian for browser detection and expose it in the language selector alongside the existing locales.

## Description

Bulgarian(BG) translation of Kite-GC.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring / Code health
- [x] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] `cargo check` (in `src-tauri`) passes
- [x] I have tested the changes locally (dev + relevant build)
- [x] Documentation (ROADMAP / CHANGELOG) updated if needed
- [x] No unrelated changes

## Notes for reviewer

Nothing other than string changes.

PR #61, opened 2026-08-19
fix(i18n): Bulgarian translations, minor improvements

teodoryantcheff:
## Description

Minor improvements to the Bulgarian translation.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

## Notes for reviewer

none


teodoryantcheff:
BTW, and it may be a stupid question, but how does one get to the the RC setup section of the app ? 
I see the strings in the json, but have no idea where they are used.
TIA.

b14ckyy:
hey For RC Control you have to enable it in the settings. then the Panel appears
image

its off by default

b14ckyy:
Thanks Teo — checked this one through the same way as #54: full structural pass (no keys
added or removed, all ICU placeholders identical to `en.json`, all values strings), CI green
on all three runners.

The style change behind it — verbal noun → imperative, and formal → informal — is applied
consistently and is the right call for buttons and menu actions. Two things I'd like fixed
before merge, both look like slips rather than choices:

1. **`logbook.openVehicle`: „Отвари апарата" → „Отвори апарата".** The imperative of „отварям"
   is „отвори"; „отвари" is the plural of the noun „отвара". Your own neighbouring strings have
   it right and are untouched by this PR — `openBattery` („Отвори тази батерия…") two lines up
   and `openMission` („Отвори тази мисия…") just below.

2. **`logbook.batteryUnlink`: „Прекъсване свръзаната батерията".** Three separate issues here:
   „свръзаната" is a typo for „свързаната"; „свързаната батерията" carries the definite article
   twice (only the first element of the phrase takes it); and „Прекъсване" needs its preposition
   („Прекъсване **на връзката с**…"). The previous value was correct on all three. Matching the
   imperative style of this PR, „Прекъсни връзката с батерията" would fit.

Three judgement calls where you have the final say as the native speaker:

- **`video.framerate`:** „Честота на кадрите" → „Честота на опресняване". The EN source is
  "Frame rate" and this labels a camera capture setting, not a display refresh rate — I'd keep
  the old term.
- **`rc.method.buttonSet`:** „Комплект бутони" → „Повече от един". "Button Set" is the method
  where several buttons each write a fixed value (see `rc.buttonSetButtons`, „Бутони (натисни →
  задай стойност)"). "More than one" drops that meaning, and it's the only entry in the list
  that no longer names its mechanism. Intentional?
- **`rc.modeAdjust`:** „Променящ" as the counterpart to „Абсолютен" — would „Относителен" or
  „Променлив" read better?

Unrelated to this PR, for whenever you next touch the file: `debug.vidMjpegBuffer` is missing
from `bg.json` (1734 vs 1735 keys). It arrived with the MediaMTX video work after your
translation, so nothing here caused it — and it falls back to English, so it's not a blocker.

Push over the two fixes and this is good to merge from my side.

teodoryantcheff:
`logbook.openVehicle`, `logbook.batteryUnlink` - done. Thanks for the lesson in Bulgarian :)
 
`video.framerate` - sounds better like it is.
 `rc.method.buttonSet` - does not have a direct literal (or a good enough) translation. "Повече от един" literally means "More than one" which captures the spirit of "a set of buttons" decently well. What you (or Claude) said -- "..._where several buttons_..."
`rc.modeAdjust` - "променящ" seems to me the better translation, given that I get what the feature actually does, that is.

PR #111, opened 2026-09-05
feat(android): background telemetry — foreground service, notification, track resync; High-Resolution 3D; off-centre inset render fix

b14ckyy:
Three commits, all device-tested on the Sony XQ-CT54 (Android 14) unless noted. Concept and decisions: Marc, 2026-09-05 (Dev-Docs `active/BACKGROUND_TELEMETRY.md`). **iOS is out of scope.**

## Background telemetry (Android)

A live link now starts a **foreground service** (`TelemetryService`, type `connectedDevice`, `START_NOT_STICKY`) with a persistent notification, so minimising Kite or switching apps no longer lets Android freeze or reap the process — link, flight recorder and track keep running; the user's disconnect and a lost link stop it.

The notification is fed from **Rust**, not the WebView (which Chromium throttles when hidden): new `link_status` aggregator — vehicle (craft name, fallback variant + board), armed / flight mode, battery (% and/or V, whatever the link carries), distance to home (FC home where a protocol reports one, else the first fix after the arm edge) — refreshed at 1 Hz only when the text changed. Fed by one-liners next to the recorder feeds in all three protocol paths; deliberately not inside the recorder, which only exists while logging is on.

**Track resync** as the safety net: Rust buffers the current flight's points (5 m spacing, cleared on the arm edge, capped); `telemetry_track_since` hands them out and the page merges them into `liveTrack` and the map trail on `visibilitychange → visible`. Screen off / Doze = an ordinary lost link (no wake lock): the recording-interrupted prompt applies when a log was live, otherwise one reconnect attempt when the page comes back — no loop.

Android 13+ asks for POST_NOTIFICATIONS on the first connect (refused → the service runs unseen). `CHANGE_NETWORK_STATE` is the Android 14 prerequisite of the `connectedDevice` type for non-BLE/USB links. `-keep class com.kitegc.app.LinkService` — the first device build had no service at all: R8 had stripped the JNI-only methods (`NoSuchMethodError`, logcat only), the trap every other bridge already has a rule for.

**Verified:** foreground service up (`isForeground=true types=CONNECTED_DEVICE`), notification following the values (`Armed · CRUISE · 12.5 V · Home 247 m`), process and values alive through 3.5 min minimised and two screen-off rounds (8 + 9 min, Kite in front and behind) with a simulated LTM link over UDP (`tools/ltm_sim.py`, new) and a real MSP link over BLE/USB. Caveat: on USB = charging = no Doze idle; the battery-powered long case is still open.

Also: the Rust log is mirrored to logcat on Android (`RustStdoutStderr`) — adb cannot read the app-private log file of a release build.

## High-Resolution 3D (all platforms)

Cesium's default renders the globe at CSS-pixel resolution regardless of DPR — native on a DPR-1 desktop, a third of the pixels on the 2.6× Sony. New setting under Power Saving: native density (on) or half of it (off, `resolutionScale` 0.5). Default on for desktop, off for phones and tablets; live switch. Measured on an iGPU desktop with video: **30 fps native → 55 fps half**. Five locales + settings reference.

## Off-centre inset really renders now

Cesium 24's `PerspectiveFrustum.clone()` drops `xOffset`/`yOffset`, and the scene renders through a clone of the camera frustum — the D16 inset reached culling but not the picture: follow/orbit cams stayed screen-centred, terrain at the left edge was culled into black wedges on the phone. `patchFrustumCloneForOffsets()` carries the offsets in `clone()`. *Regression: introduced by the phone chrome (PR #103), never released.*

## 3D mission markers

Rastered at ⌈DPR⌉× their intrinsic size (Cesium draws billboards in CSS px — a 48 px icon was 126 blurry device px) and sized with the 2D marker factor 0.85, glow and anchor offsets included. Not yet eyeballed on the phone.

## Checks

`npm run check` 0/0, `cargo check`, `cargo test` (new unit tests for the buffer gating and the notification text), local arm64 release APK. The Android-only Rust and the Kotlin service compile only in the Android build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


b14ckyy:
3D mission markers eyeballed on the Sony: OK (still on the large side by design — readable outdoors at arm's length).

PR #114, opened 2026-09-06
fix: macOS signing, location permission and native CoreLocation, GCS marker fallback, toast de-dup

SebastianKumor:
## Description

Five fixes found while testing the macOS end of 1.0.0. Each is its own commit and they are
independent, so any subset can be taken. Opened as a draft against `master` rather than
`development`, because the first one only affects the signed release artifact and the second and
third only affect macOS.

**1. Notarized builds shipped with an empty entitlement set** (`scripts/notarize-macos.sh`)

`codesign --force` replaces the bundler's signature, and codesign only embeds the entitlements it is
explicitly handed. Signing without `--entitlements` produced a hardened-runtime binary that declares
nothing, so the runtime denied everything `Entitlements.plist` asks for: no CoreBluetooth central for
the BLE transport, no device for the "Camera (device)" video source, no CoreLocation. Unsigned local
builds are unaffected, which is why this never shows up in testing.

Verified by signing a bundle both ways and dumping the result:

```
--force --deep --options runtime --sign ...          ->  no entitlements embedded
same + --entitlements src-tauri/Entitlements.plist   ->  bluetooth, camera, location all present
```

**2. macOS never declared the location permission** (`src-tauri/Info.plist`, `Entitlements.plist`)

No `NSLocationWhenInUseUsageDescription` and no location entitlement, so macOS would not authorise
CoreLocation and could not even show a consent prompt. This is the prerequisite for commit 5.

**3. Continuous GCS mode discarded a position it already had** (`src/lib/stores/gcsLocation.ts`)

`recompute()` handled only `off` and `manual`, so `continuous` relied entirely on its location watch.
Where that watch cannot deliver, the marker stayed null even though `userGeoLocation` held a usable
position: a one-shot OS fix, the value persisted from last session, or the connected UAV's own GPS
fix that `helpers/userLocation.ts` captures once per connection. `continuous` is the default, so the
app knew where the operator was and showed nothing. Going from `off` to `continuous` had the same
effect, since `off` clears the marker and nothing re-seeded it.

`recompute()` now also seeds `continuous` from the last known position while the watch has produced
nothing, and runs on every mode change. The first real fix overwrites the seed and a delivered
position is left alone, so live tracking is unchanged wherever it already worked.

**4. Toast de-dup only compared the last line** (`statusText.ts`, `StatusTextToasts.svelte`)

An FC nagging with two alternating messages never matched. INAV does exactly that while waiting for a
fix ("UNABLE TO ARM" / "WAITING FOR GPS FIX"): neither is ever the previous line, so every nag
appended a row, the banner hit its 12-line cap within seconds, and the tone re-cued each time. A
repeat now refreshes the matching line wherever it sits and bumps a dimmed `xN` counter, keeping its
position so the banner does not reshuffle while being read.

**5. Native CoreLocation source for the GCS marker** (new `src-tauri/src/location_macos.rs`)

macOS is the only desktop platform with no Web Geolocation API, so `navigator.geolocation` could
never resolve there and the marker had no source at all. WKWebView ships no Geolocation API (wry
implements the permission prompt only for its Android backend, WebKit exposes no delegate), and
`tauri-plugin-geolocation`'s desktop backend is a stub returning a default position. Windows
(WebView2) and Linux (WebKitGTK) support the Web API and mobile uses the native plugin, so only macOS
needed this.

Adds a `CLLocationManager` plus delegate in Rust via objc2, the same in-crate approach `hid/macos.rs`
already uses for its C API: no Swift, no extra plugin. `helpers/userLocation.ts` routes macOS to it
and subscribes to an `os-location` event, which is what lets `continuous` track without a
`watchPosition`. Every other target compiles a stub so `generate_handler!` needs no platform branch.

Confirmed working on a real Mac: permission prompt appears, marker resolves, `Continuous (live)`
tracks.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Documentation
- [x] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

`npm run check`: 559 files, 0 errors, 0 warnings. `cargo check`: clean. Release bundle rebuilt and
installed; `otool -L` confirms CoreLocation linked and Contacts absent.

No changelog entry: the 1.0.0 section is a prose summary with no itemised fix list, so there is
nowhere natural for these. Happy to add entries if you would rather have them.

## Notes for reviewer

- Written directly against `master`, not cherry-picked, so nothing from `development` rides along.
  Fixes 2 and 4 also exist there (the plist key via #16, the de-dup via earlier work), so expect a
  trivial conflict or a no-op when master next merges up.
- Commits 3 and 4 are cross-platform, not macOS-only. Commit 3 changes when `recompute()` runs for
  `continuous` on every platform, which is the part most worth your eyes even though live tracking
  should be unaffected.
- Commit 5 is the only one adding a dependency: `objc2-core-location 0.3.2`. `objc2 0.6` and
  `objc2-foundation 0.3` were already in the tree via tauri, wry and tao, so no second objc2
  generation enters the binary. Taken with `default-features = false` because the default set pulls
  in `objc2-contacts` for `CLPlacemark` geocoding. Checked against the RustSec advisory database: no
  advisories for any of the three.
- Commit 5 is a feature rather than a fix, so if you want `master` strictly frozen it is the one to
  drop; commits 2 and 3 then still leave the marker working off the vehicle's GPS fix.
- Three details in `location_macos.rs` worth knowing before touching it: the manager is created on
  the main thread because it needs a live run loop, `delegate` is a WEAK property so the manager and
  delegate are parked in a process-global, and a denial or disabled Location Services only logs so
  start-up can never block.
- All five bugs are present in `v1.0.0-rc2` and earlier, confirmed against the tag.

### DIFF
--- a/scripts/notarize-macos.sh
+++ b/scripts/notarize-macos.sh
@@ -45,7 +45,14 @@ fi
 
 echo "[1/5] Code-signing the app bundle (hardened runtime)..."
 # --options runtime is required for notarization; --deep signs nested frameworks/helpers.
+# --entitlements is NOT optional here: codesign only embeds the entitlements it is handed, and
+# --force replaces whatever signature the bundler left behind. Signing without it produces a
+# hardened-runtime binary with an EMPTY entitlement set, and the hardened runtime then denies the
+# resources Kite declares in Entitlements.plist: the BLE transport cannot open a CoreBluetooth
+# central, the "Camera (device)" video source gets no device, and the GCS marker gets no location.
+# None of that reproduces in an unsigned local build, so it only ever breaks the released .dmg.
 codesign --force --deep --options runtime --timestamp \
+    --entitlements "$ROOT/src-tauri/Entitlements.plist" \
     --sign "$APPLE_SIGNING_IDENTITY" "$APP"
 
 echo "[2/5] Signing the .dmg..."
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2468,6 +2468,9 @@ dependencies = [
  "log",
  "mavlink",
  "num-traits",
+ "objc2 0.6.4",
+ "objc2-core-location",
+ "objc2-foundation 0.3.2",
  "reqwest 0.12.28",
  "rusqlite",
  "serde",
@@ -3007,6 +3010,16 @@ dependencies = [
  "objc2 0.6.4",
  "objc2-core-foundation",
  "objc2-io-surface",
+]
+
+[[package]]
+name = "objc2-core-location"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ca347214e24bc973fc025fd0d36ebb179ff30536ed1f80252706db19ee452009"
+dependencies = [
+ "objc2 0.6.4",
+ "objc2-foundation 0.3.2",
 ]
 
 [[package]]
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -83,4 +83,17 @@ webkit2gtk = "2.0"
 # used to build the device-matching dictionaries and read device properties.
 core-foundation = "0.10"
 core-foundation-sys = "0.8"
+# Native CoreLocation for the GCS marker (location_macos.rs). WKWebView ships no Geolocation API at
+# all (wry implements the permission prompt only for its Android backend, and
+# tauri-plugin-geolocation's desktop backend is a stub that returns a default position), so
+# `navigator.geolocation` can never resolve on macOS and the marker had no source but the vehicle's
+# own GPS. We drive CLLocationManager through the objc2 bindings instead, in-crate and without Swift,
+# the same way the IOKit HID backend above talks to its C API. Only the features for the types used.
+# The 0.6 / 0.3 generation is the one tauri, wry and tao already pull in, so this adds no second
+# objc2 to the binary (btleplug's older 0.5 copy is pre-existing and unrelated).
+objc2 = "0.6"
+objc2-foundation = { version = "0.3", features = ["NSArray", "NSString", "NSEnumerator"] }
+# default-features = false on purpose: the crate's default set drags in objc2-contacts (CNPostalAddress,
+# for CLPlacemark geocoding) which a GCS has no business linking.
+objc2-core-location = { version = "0.3", default-features = false, features = ["std", "CLLocation", "CLLocationManager", "CLLocationManagerDelegate", "CLError"] }
 
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -92,7 +92,7 @@ core-foundation-sys = "0.8"
 # The 0.6 / 0.3 generation is the one tauri, wry and tao already pull in, so this adds no second
 # objc2 to the binary (btleplug's older 0.5 copy is pre-existing and unrelated).
 objc2 = "0.6"
-objc2-foundation = { version = "0.3", features = ["NSArray", "NSString", "NSEnumerator"] }
+objc2-foundation = { version = "0.3", features = ["NSArray", "NSString", "NSEnumerator", "NSError"] }
 # default-features = false on purpose: the crate's default set drags in objc2-contacts (CNPostalAddress,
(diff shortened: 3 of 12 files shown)


b14ckyy (changes requested):
Thanks, this is a good batch. Checked on the PR head here (Windows): `npm run check` 0/0, `cargo check` and `cargo test --no-run` clean, the objc2 deps sit in the macOS target block as they should. Short version per commit: the entitlements fix is the real find (the bundler config in tauri.conf.json had them, `--force` threw them away again), the plist/entitlement keys are fine, the de-dup is byte-identical to what `development` already carries so the merge-up is a no-op, and the CoreLocation module reads correctly (weak delegate parked in the static, main-thread hop, denial only logs). I'll take all five. Two things to fix before I merge:

**1. `continuous` on macOS is a one-shot seed, not live tracking**

`recompute()` only seeds the marker while `gcsLocation` is null:

```ts
} else if (mode === 'continuous' && !get(gcsLocation)) {
  gcsLocation.set(get(userGeoLocation));
```

On macOS the WebView watch in `startContinuous()` never delivers, so the marker gets set exactly once, by the first CoreLocation fix. Every later fix still arrives (the `os-location` event goes through `setUserLocation` into `userGeoLocation`, and `userGeoLocation.subscribe(() => recompute())` fires), but `recompute()` then sees a non-null marker and returns without touching it. So the marker follows the operator until the first fix and freezes there; "Continuous (live) tracks" holds for the first position only. Easy to miss in a test at a desk, since the first fix and "tracking" look the same until you move more than 20 m.

Suggested fix, in line with what the commit message already describes: remember whether the Web watch has ever delivered (a `watchDelivered` flag set in the `watchPosition` callback, cleared in `clearWatch()`), and while it has not, let `continuous` follow `userGeoLocation` with the same 20 m gate the watch uses:

```ts
} else if (mode === 'continuous' && !watchDelivered) {
  const next = get(userGeoLocation);
  const cur = get(gcsLocation);
  if (next && (!cur || haversineDistance(cur.lat, cur.lon, next.lat, next.lon) > CONT_MIN_MOVE_M)) {
    gcsLocation.set(next);
  }
  gcsAccuracyM.set(get(userGeoAccuracyM));
}
```

That keeps Windows/Linux unchanged once their watch is live (the flag flips on the first fix), and on macOS the native module's own 20 m distance filter already keeps the event stream quiet. The UAV-GPS fallback is captured once per connection, so nothing thrashes.

**2. "Detect my location" on macOS never restarts CoreLocation**

`location::start()` bails out for good when Location Services are off system-wide at launch, and the settings button only calls `location_os_last`, which then stays `None` forever. A user who enables Location Services (or grants the permission) after launch has to restart the app to get a marker. Please have the button path call `start()` again: expose it as a command (`location_os_start`, stub on other targets) and invoke it from `runGeoCheckMacOs()` before reading the last fix. `start()` is already idempotent, so this costs nothing on the happy path, and the `locationServicesEnabled_class()` check turns from a permanent give-up into a per-click retry.

Minor, take it or leave it: `objc2-foundation` uses `NSError` but does not list the `NSError` feature; it only compiles because tauri/wry enable it. One entry in the feature array makes that explicit.

Once these two are in and your signed 1.0 build is redone on top of the merged notarize script, we are good to publish.

SebastianKumor:
Both fixed, plus the `NSError` feature, in 0ed7aa0 / ead10a9 / 1dc6e1a. `npm run check` 0/0, `cargo check` clean.

You are right about the freeze, and my "Continuous (live) tracks" claim was wrong: I only ever saw the first fix land. Now gated on whether the watch has delivered, not on an empty marker, so Windows/Linux are untouched once their watch is live.

One caveat before you merge: I can confirm the path runs, but not that it tracks, since that needs the Mac to physically move more than 20 m. Taking the laptop outside next and will confirm here.

b14ckyy:
No need to run outside with the MacBook. It probably does not have GPS and you would need to get into other WiFi areas that are mapped to actually change location. if the code confirms that it keeps tracking, I am fine with that and if it ever gets reported as non working, there can always be a 1.0.1 patch. I doubt the continuous tracking will be used by many on a desktop system in a driving car. 

So if you are good, then set it to ready and I will merge the PR. 

There will be 2 more PRs from me in a bit and then  I think we are done.

b14ckyy (approved):

PR #131, opened 2026-09-07
fix(control): Fly Here on INAV over MAVLink, MAVLink as the default protocol, INAV HDOP

SebastianKumor:
## Description

Guided control ("Fly Here") did not work on INAV over MAVLink, and two smaller bugs turned up while
tracking that down. All of it was found and verified against a flight-controller simulator, which is
included here so the paths can be tested without hardware.

**Fly Here on INAV over MAVLink.** INAV's `handleIncoming_COMMAND_INT` accepts `DO_REPOSITION` only
with `MAV_FRAME_GLOBAL`; the relative-alt and terrain frames are commented out in the firmware, so
it answers `MAV_RESULT_UNSUPPORTED`. Kite sends `MAV_FRAME_GLOBAL_RELATIVE_ALT`, which is the right
frame for ArduPilot and the meaning of the altitude in the UI. The reposition now retries once in
`MAV_FRAME_GLOBAL` with an AMSL altitude, and only on that specific refusal, so ArduPilot's first
attempt is accepted and it never sees a second frame. The AMSL offset comes from
`GLOBAL_POSITION_INT`, which carries both altitudes: INAV never sends `HOME_POSITION`, so there is
no home altitude to work from on that link.

**MAVLink is the default protocol.** MSP was the desktop default, but the common ground station is a
telemetry radio, which is MAVLink over serial, and MAVLink is the only protocol the control panel
supports. The network port now defaults per protocol as well (MAVLink UDP 14550 or TCP 5760, MSP TCP
5761) instead of forcing TCP to the INAV SITL MSP port whichever protocol was selected; a port typed
by hand is left alone. Restoring a saved protocol no longer falls back to MSP for an unrecognised
value, which also stops it rewriting a stored `telemetry` into `msp`.

**INAV HDOP was the wrong field.** `MSP_GPSSTATISTICS` starts with `lastMessageDt`, a `uint16` in
`fc_msp.c`. Decoding it as 32 bits shifted every later field by two bytes, so the GPS tile showed
`eph` as HDOP, and `eph`/`epv` were off by two in the recorder too.

**Fly Here radius.** On fixed wing the field opened blank, which reads as "no radius" while the
vehicle would use its configured one; it is now seeded from `WP_LOITER_RAD`, already read for the
loiter ring. The popup was 240 px while the two steppers need ~274 px, so the radius "+" was clipped.

**Simulator (`tools/fc_sim.py`).** One process, three combinations: ArduPilot over MAVLink, INAV over
MSP, and INAV on its own MAVLink port. Commands change real vehicle state rather than being
acknowledged and ignored, which is the failure a control test has to be able to catch. MAVLink
layouts, CRC_EXTRA seeds and enums are read at runtime from the dialect XML the app already links, so
they cannot drift; MSP layouts follow `fc_msp.c` field order. Both airframes fly to their own
firmware defaults, taken from ArduPlane master (`config.h`) and INAV master (`settings.yaml`), and
the INAV MAVLink mode reproduces that port's real limits: `COMMAND_LONG` has no handler and is
dropped without an ACK, `COMMAND_INT` handles only `DO_REPOSITION`, guided needs `isGCSValid()`,
missions use `MISSION_ITEM` rather than the `_INT` variant, and `PARAM_REQUEST_LIST` answers empty.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

## Notes for reviewer

Testing, all against the included simulator on a signed local build:

| Check | Result |
| --- | --- |
| `cargo test` | 157 passed, 0 failed |
| `npm run check` | 577 files, 0 errors, 0 warnings |
| ArduPilot MAVLink harness | 30 checks pass |
| INAV MAVLink harness | 34 checks pass |
| INAV MSP harness | 34 checks pass |
| Reposition frame fallback | first attempt UNSUPPORTED, retry ACCEPTED, aircraft flies to the target at the commanded altitude |
| Waypoint sequencing | 4 course/firmware combinations |

Fly Here was also confirmed by hand in the app against the INAV MAVLink simulator.

Points worth your attention:

1. **The default protocol change affects every existing user**, though only on a fresh profile: a
   stored `lastProtocol` still wins. This is the one change here that is a judgement call rather
   than a fix, so say the word and I will drop that commit.
2. **`wait_for_ack` is split** into a version returning the raw `MavResult` plus a formatter. The
   retry has to know *which* refusal it got, and matching on the human-readable error string would
   break the moment that wording changed.
3. **The AMSL offset lives in the MAVLink layer**, not the frontend. My first version computed it as
   `altMsl - altitude` in `vehicleControl.ts`, which is wrong: `altMsl` arrives from `GPS_RAW_INT`
   and the relative altitude only from `GLOBAL_POSITION_INT`, so a GPS-first ordering produced an
   offset the size of the whole AMSL altitude. It is cleared when a handler starts so a new vehicle
   cannot inherit the last one, and there is a unit test for both.
4. **What this does not fix:** on INAV over MAVLink the Guided button, the mode buttons, RTL, Abort
   Landing, Change speed, Set Loiter Radius and Set Home Here all travel as `COMMAND_LONG` and are
   silently dropped by the firmware, so they time out. Fly Here and Change Alt are the two that use
   `COMMAND_INT`. Hiding or relabelling the rest when the FC is INAV is a separate question, partly
   because Kite cannot currently tell INAV from ArduPlane over MAVLink: with
   `mavlink_autopilot_type = ARDUPILOT` INAV reports `MAV_AUTOPILOT_ARDUPILOTMEGA` and ArduPlane
   mode numbers, and it never sends `AUTOPILOT_VERSION`. The reliable tell is that ArduPilot answers
   `AUTOPILOT_VERSION` and serves parameters while INAV answers neither. Happy to do that separately
   if you want it.
5. **Also missing from Kite's plane mode table:** `AUTOLAND = 26`, which exists in ArduPlane master.
   Not touched here to keep this focused.


b14ckyy (changes requested):
Thanks for the thorough work, Sebastian, and for the simulator in particular. I went through the PR against the INAV sources, and I would like to take it apart rather than merge it as one piece. Here is what happens to each part.

**Goes in, as one patch PR against `release/1.0.x`**

I have just cut `release/1.0.x` from the `v1.0.0` tag; that is where 1.0 maintenance lives now (`master` has moved on to 1.1 snapshots). Please rebase these three onto it and open one PR with base `release/1.0.x`:

1. The HDOP fix. Verified: `MSP_GPSSTATISTICS` in `fc_msp.c` writes `lastMessageDt` as `U16` on 7.0.0 and on master, so HDOP is at offset 14 and Kite was reading `eph`. The layout has not moved since before 7.0 (master only appends `hwVersion`), so no version gating is needed.
2. The Fly Here radius seed from `WP_LOITER_RAD` and the popup width.
3. From the connection commit: only the restore fix, i.e. a saved `telemetry` protocol is honoured instead of being rewritten to `msp`. Note that on the 1.0 tree the transport/port toggle still lives in `Toolbar.svelte` (`ConnectionControls.svelte` does not exist there yet). MSP stays the default, on every platform; see below.

Two conventions for that PR: put `Affects: v1.0.0 and earlier` in the description (every fix states whether it was in the last release or is a regression since), and the changelog paragraphs for these three move into it, worded as fixes.

**Not going in**

*MAVLink as the default protocol.* I would like to keep MSP as the default, and I will be honest that this is a personal priority rather than a technical argument: I fly INAV, and that is what I reach for first. Kite started with INAV, but it is not meant to be an INAV-first ground station; the goal is to treat every stack as equally as we can, and the default is simply the one thing that has to pick a side. So please drop the default change itself. The per-protocol port re-default in the same commit is a good idea, but it is a behaviour change rather than a fix, so it does not belong in the patch; please bring it to `development` for 1.1 instead, with MSP/5761 staying the defaults. What I would also take there: a setting to hide protocols you never use (a checkbox per protocol), so someone who never touches MSP does not see it in the selector at all. That helps your use case without moving the default for everyone.

*The INAV `DO_REPOSITION` fallback.* I would rather not take this, and not grow INAV-specific behaviour on the MAVLink path for now. INAV's supported path in Kite is MSP; INAV over MAVLink works for telemetry, but control over that port is something I do not want to promise until the firmware side is steadier. The findings that led there, so the reasoning is on record:

- `handleIncoming_COMMAND_INT` exists only since INAV 9.0.0 (`mavlink_gcs_nav`, first commit d09af59). 7.x and 8.x do not handle `COMMAND_INT` at all, so they never ACK and the retry never fires. The feature would be 9.x-only, and Kite cannot tell over MAVLink: INAV sends no `AUTOPILOT_VERSION` and reports as ArduPilotMega. Without a way to gate by firmware version there is no safe, reliable way to ship it.
- The altitude is wrong for the firmware that is out there. INAV insists on `MAV_FRAME_GLOBAL`, but hands `z` to `setWaypoint(255, …)`, which converts with `GEO_ALT_RELATIVE` (`navigation.c` ~3967 on 9.1.0 and master; `navigation_geo.c:129`, "llh altitude is already relative to origin", `pos->z = llh->alt`). INAV takes the value as metres above home regardless of the frame. The PR adds the field elevation before the retry, so a real board would fly to `target + field elevation` above home: an 80 m target on a 300 m field becomes 380 m AGL. The simulator agrees with the PR only because its INAV model implements the MAVLink spec (`alt - v.home_alt`) rather than INAV's code, so the test validated its own assumption.
- INAV's handler carries the other frames commented out and a `// TODO: Alt modes`. That part of the firmware is still moving, and I am not comfortable adapting Kite's command chain to it while it does. Once INAV settles this upstream we can look again; the frame/altitude mismatch is worth reporting to them, and I would welcome you doing that since you have the analysis fresh.

The same picture holds for missions on that port, which is the other half of what a control path would need. `handleIncoming_MISSION_ITEM` on master accepts exactly two commands, `NAV_WAYPOINT` and `NAV_RETURN_TO_LAUNCH`, answers `MAV_MISSION_UNSUPPORTED` to everything else, and zeroes `p1`/`p2`/`p3` on the way in, so no per-waypoint speed, hold time, altitude mode or user action ever reaches the FC. Most of what ArduPilot or PX4 missions carry is either rejected or flattened to a plain position. Even if Kite had a dedicated INAV mission profile for that port (it will not), it could not switch to it automatically, because over MAVLink it cannot tell that the vehicle is INAV, let alone which version; and every INAV release that touches that port would have Kite chasing it. That is the wider reason for keeping INAV on MSP, where the protocol is versioned and the feature set is gated properly.

**The simulator**

`tools/fc_sim.py` is worth keeping for the future: stdlib only, self-contained, and the MSP side is a real asset for hardware-free testing. It does not need a PR; it can land as a direct commit on `master` under `tools/`, authored by you, since it is a dev tool and counts as documentation for our purposes. Either open a tools-only PR against `master` or say so and I will commit it with your authorship. One request either way: a short note at the top of the INAV MAVLink model that it implements the spec's `MAV_FRAME_GLOBAL` semantic while the firmware treats the altitude as relative to home, so nobody validates against it again without knowing.

So: one patch PR against `release/1.0.x` with the three fixes, the simulator to `master`, the port re-default and the protocol-hiding setting as a 1.1 PR against `development` if you want to pick them up, and the rest closed with this PR.

SebastianKumor:
Understood on the default, and dropping it. Counter-proposal for 1.1 instead of moving it: a **Default protocol** setting in global settings, persisted, shipping as MSP so nothing changes for anyone who does not touch it.

That composes with your per-protocol hide checkboxes: a hidden protocol simply cannot be the default. Worth noting the last-used protocol is already restored on start, and the restore fix in this PR is what makes that reliable for Telemetry, so the setting is really about the first run and about not letting last-used silently decide.

Happy to build both against `development`, together with the per-protocol port re-default.

SebastianKumor:
Split up as you asked, so closing this in favour of:

- **#143** against `release/1.0.x`: the three accepted fixes, `Affects: v1.0.0 and earlier`, with the changelog paragraphs moved in and reworded as fixes. Fix 3 is reworked for that tree, since `ConnectionControls.svelte` does not exist there.
- **#142** against `master`: the simulator, with the altitude note you asked for.
- **iNavFlight/inav#11884**: the frame/altitude mismatch reported upstream, with both possible fixes and an offer to PR whichever they prefer.

Dropped: the default-protocol change, and the `DO_REPOSITION` fallback. You were right about the altitude on that one, and I verified it before accepting it: `setWaypoint(255, …)` converts with `GEO_ALT_RELATIVE`, which is `pos->z = llh->alt` verbatim, so the retry would have flown target-plus-field-elevation above home. My simulator agreed with my own code because its INAV model implemented the spec rather than the firmware, so the test confirmed its own assumption. That is fixed in #142 and the check now demonstrates the trap instead.

The port re-default and the default-protocol setting are noted for a 1.1 PR against `development` whenever you want them.

b14ckyy:
> Understood on the default, and dropping it. Counter-proposal for 1.1 instead of moving it: a **Default protocol** setting in global settings, persisted, shipping as MSP so nothing changes for anyone who does not touch it.
> 
> That composes with your per-protocol hide checkboxes: a hidden protocol simply cannot be the default. Worth noting the last-used protocol is already restored on start, and the restore fix in this PR is what makes that reliable for Telemetry, so the setting is really about the first run and about not letting last-used silently decide.
> 
> Happy to build both against `development`, together with the per-protocol port re-default.

I like that Idea! A drop down that is set to "Last" by default. Existing behavior. But can be set to MSP/Mavlink/Telemetry so on every app launch it sets to that value by default (overrides the last used store). I am fine with that.

PR #143, opened 2026-09-09
fix: INAV HDOP field offset, Fly Here radius field, saved Telemetry protocol restore

SebastianKumor:
## Description

The three fixes from #131 that you accepted, rebased onto `release/1.0.x` as one patch PR.

**Affects: v1.0.0 and earlier.** None of these are regressions; all three are in the released build.

1. **HDOP read the wrong field.** `MSP_GPSSTATISTICS` starts with `lastMessageDt`, which INAV packs
   as a `uint16` (`src/main/fc/fc_msp.c`). Decoding it as 32 bits shifted every later field by two
   bytes, so the GPS tile showed `eph` as the HDOP, and `eph`/`epv` were off by two in the recorder.
   As you confirmed, the layout has not moved since before 7.0 and master only appends `hwVersion`,
   so there is no version gating.
2. **Fly Here opened with a blank, clipped radius field** on fixed wing. It now seeds from the FC's
   own `WP_LOITER_RAD`, already read into `fcLoiterRadius` for the loiter ring, and an explicit
   value from the session still wins. The popup went from 240 px to 300 px, because each
   NumberStepper is a fixed ~132 px and the alt + radius pair needs ~274 px, so the radius "+" was
   cut off by the popup edge.
3. **A saved Telemetry protocol came back as MSP.** The restore mapped everything that was not
   `mavlink` onto `msp`, silently rewriting a stored `telemetry` choice. MSP stays the fallback for
   an unrecognised or missing value, so a fresh install is unchanged.

Nothing else from #131 is here: the default-protocol change is dropped, and the per-protocol port
re-default waits for a 1.1 PR against `development` as you asked.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

## Notes for reviewer

Per your note, fix 3 is reworked for this tree: `ConnectionControls.svelte` does not exist on 1.0.x,
so the restore lives in `+page.svelte`. The transport/port toggle in `Toolbar.svelte` is untouched
and MSP is still the default everywhere.

Verification: `cargo check` clean, `npm run check` 559 files 0 errors 0 warnings, and the HDOP fix
re-checked against a simulated FC sending the firmware layout (hdop 1.10, eph 1.50): the tile read
1.5 before and reads 1.1 now.

Three things for you to decide rather than me:

1. **Changelog placement.** I added a `1.0.1 — patch release` section with the four paragraphs
   worded as fixes, below the `1.1.0 — in development` block that is still present on this branch.
   That 1.1 block looks like a leftover from the tag cut; removing it felt like your call, not
   something to slip into a patch PR.
2. **The `[#PRNUM]` placeholders** in those changelog entries. Tell me this PR number is fine to
   use and I will amend them to it, or say if you would rather they carry no reference on a patch
   branch.
3. **A stray debug print in the released build**: `src-tauri/src/scheduler/telemetry.rs` has an
   `eprintln!("[GPS-STATS] …")` that fires on every GPS statistics message. I left it alone to keep
   this to the three approved fixes; happy to drop it here or in its own PR, whichever you prefer.

### DIFF
--- a/docs/user/changelog.md
+++ b/docs/user/changelog.md
@@ -108,6 +108,22 @@ below. The release you are reading the docs for is expanded; click an older vers
       and after any packet loss the video pauses until the next keyframe instead of freezing the
       Pi's hardware decoder. Kernel-side report: raspberrypi/linux#7609. [#112]
 
+??? note "1.0.1 — patch release"
+
+    **Fixed**
+
+    - **HDOP on INAV showed the wrong figure.** The GPS tile read the position-error field instead
+      of HDOP, because the first field of INAV's GPS statistics message is 16 bits and Kite decoded
+      it as 32. The recorder stored the same two values one field out. [#PRNUM]
+    - **"Fly Here" opened with an empty radius field** on fixed wing, which reads as "no radius"
+      while the vehicle would in fact use its configured one. It now starts at the aircraft's own
+      loiter radius, and an explicit value from this session still wins. [#PRNUM]
+    - **The radius stepper was clipped** by the edge of the "Fly Here" popup, so its "+" button
+      could not be reached. [#PRNUM]
+    - **A saved Telemetry connection came back as MSP.** Restoring the last-used protocol mapped
+      everything that was not MAVLink onto MSP, so the passive Telemetry choice was silently
+      rewritten. [#PRNUM]
+
 ??? note "1.0.0 — Initial release"
 
     The first stable release of **Kite Ground Control**: a cross-platform ground station for
--- a/docs/user/changelog.md
+++ b/docs/user/changelog.md
@@ -114,15 +114,16 @@ below. The release you are reading the docs for is expanded; click an older vers
 
     - **HDOP on INAV showed the wrong figure.** The GPS tile read the position-error field instead
       of HDOP, because the first field of INAV's GPS statistics message is 16 bits and Kite decoded
-      it as 32. The recorder stored the same two values one field out. [#PRNUM]
+      it as 32. The recorder stored the same two values one field out. [#143]
     - **"Fly Here" opened with an empty radius field** on fixed wing, which reads as "no radius"
-      while the vehicle would in fact use its configured one. It now starts at the aircraft's own
-      loiter radius, and an explicit value from this session still wins. [#PRNUM]
+      while the vehicle would in fact use its configured one. The field now shows the aircraft's own
+      loiter radius. Leave it alone and the aircraft keeps using that setting, turn direction
+      included; type a value and yours wins. [#143]
     - **The radius stepper was clipped** by the edge of the "Fly Here" popup, so its "+" button
-      could not be reached. [#PRNUM]
+      could not be reached. [#143]
     - **A saved Telemetry connection came back as MSP.** Restoring the last-used protocol mapped
       everything that was not MAVLink onto MSP, so the passive Telemetry choice was silently
-      rewritten. [#PRNUM]
+      rewritten. [#143]
 
 ??? note "1.0.0 — Initial release"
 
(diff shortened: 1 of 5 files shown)


b14ckyy (changes requested):
Thanks, Sebastian — this is the shape I asked for. I re-verified all three against the sources and the 1.0.x tree, and the checks pass here too (svelte-check 559 files / 0, `cargo check` clean, branch on the current tip).

**Verified**

- HDOP: `MSP_GPSSTATISTICS` on INAV master writes `lastMessageDt` U16, then three U32, then hdop/eph/epv U16 and `hwVersion` U8, so 14 is right and the eph/epv length guards fit both the 20-byte (pre-hwVersion) and 21-byte payloads.
- Restore: `ProtocolType` on 1.0.x is `msp | mavlink | telemetry`, the Toolbar selector offers all three, MSP stays the fallback. Minimal and correct.
- Popup: the ~132 px per stepper checks out; 300 px fits the pair and `flex-wrap` is a sensible fallback.

**One change request: the radius seed flips the loiter direction on ArduPilot**

Before, a blank field sent `param3 = 0`, and `Plane::set_guided_WP` then took radius *and* direction from `WP_LOITER_RAD` (negative = counter-clockwise). With the seed Kite always sends a positive radius (`fcLoiterRadius` stores the magnitude) and, for planes, `param4 = NaN`. `handle_command_int_do_reposition` turns that NaN into `loiter_ccw = 0` and calls `set_radius_and_direction(radius, false)`, which forces clockwise. A plane with `WP_LOITER_RAD = -80` therefore loiters clockwise on Fly Here instead of as configured. That was already true for a hand-typed radius, but the seed makes it the default path for every Fly Here on a fixed wing (PX4 most likely behaves the same; I did not verify it there).

Please keep the seed as a *display* value only: in `fly()` send `null` while the user has not touched it (no session value yet and `radius` still equals `$fcLoiterRadius`), and do not persist the seed into `guidedParams` in that case. That also avoids the seed sticking as an "explicit" session value across a reconnect to a different aircraft. Carrying the direction through `param4` is a 1.1 topic, not for the patch.

**Your three questions**

1. Changelog placement: keep the `1.1.0 — in development` block. It is not a leftover; it is in the `v1.0.0` tag, and `docs/hooks/changelog_channel.py` drops every "in development" section on the release channel and then expands the first remaining one. So `??? note "1.0.1 — patch release"` between 1.1.0 and 1.0.0, collapsed, is exactly right, and removing the block would only create forward-port conflicts.
2. Use `#143`. The file uses reference-style links, so please also add `[#143]: https://github.com/b14ckyy/Kite-GC/pull/143` to the definitions at the bottom, otherwise it renders as literal brackets.
3. The `eprintln!`: please switch it to `log::debug!` in this PR. Our convention is that `eprintln!` is dev-only and never reaches the release log file, while `log::debug!` lets a tester confirm the fix on real hardware with the log level on Debug. The file already uses `log::`, and you are touching that function anyway.

Label and milestone are set. Once the seed change is in, this is good to merge.

SebastianKumor:
You are right, and the direction was the part I missed: `fcLoiterRadius` stores the magnitude, so echoing the seed back throws the sign away. `fly()` now sends `null` while the field still shows the untouched seed, which lands as `param3 = 0` in `commands/control.rs:138`, exactly what a blank field used to send, and the seed is no longer persisted into `guidedParams` in that case.

Also done: `eprintln!` is `log::debug!`, and the four `[#PRNUM]` are `[#143]` with the link definition added. Changelog block left where it is.

`npm run check` 559 files / 0, `cargo check` clean.

b14ckyy (approved):

PR #150, opened 2026-09-10
fix(ios): keep libapp.a out of the app bundle

SebastianKumor:
## Description

The iOS archive is rejected by App Store validation:

```
Invalid bundle structure. The "Kite Ground Control.app/libapp.a" binary file is not permitted.
```

`tauri ios init` lists `Externals/` as a plain source group. XcodeGen has no default build phase for
a `.a` and falls back to `resources`, so the Rust static library is added to "Copy Bundle Resources"
as well; the link itself comes from the target's `dependencies: - framework: libapp.a`, not from the
group. The copy puts the library inside the app, which Apple refuses, and it carried the library
into the ipa for nothing: 56 MB before, 19 MB after. Once several copies exist side by side (debug +
release, device + simulator) the build fails outright with `Multiple commands produce .../libapp.a`.

`src-tauri/gen/apple/` is generated and gitignored, so the project file cannot carry the fix. The
scripts apply `buildPhase: none` to the group and then **regenerate the project**, because a patched
`project.yml` on its own reaches nothing: only `tauri ios init` runs XcodeGen, while `tauri ios
build` and `tauri ios dev` load the existing pbxproj and synchronize values into it. Regeneration
happens only when it is needed, i.e. when the patch changed the file or the existing pbxproj still
copies the library, and the signing team survives it because that lives in `project.yml`.

The plumbing is shared by `scripts/build-ios.sh` and `scripts/run-ipad.sh` in
`scripts/lib/ios-project.sh`. That matters: a dev run generates the project while `Externals/`
already holds a built library, so it produces exactly the pbxproj a later release archive is
rejected for.

Both scripts also had the same path bug: the init check and the output hint looked for `gen/apple`
rather than `src-tauri/gen/apple`, so `tauri ios init` ran on every single build and the path
printed at the end did not exist.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

## Notes for reviewer

Scripts only, no Rust or frontend change, so both checks are unaffected; I ran them anyway. No
changelog entry, since the bug never reached a release.

That `tauri ios build` does not regenerate the project is measured, not assumed: with a marker
setting added to `project.yml`, a full `tauri ios build` reached `xcodebuild` without the marker ever
appearing in `project.pbxproj`, whose mtime did not move.

`ios_prepare_project` tested against the real generated project:

| Case | Result |
| --- | --- |
| Already correct | no regeneration, nothing touched |
| Unpatched yml + pbxproj copying libapp.a (the state a dev run leaves) | patched, regenerated, 2 Resources entries to 0 |
| Patched yml + stale pbxproj | regenerated, 2 to 0 |
| Upstream's `excludes: ["**/*.a"]` form (tauri-apps/tauri#15890) | `buildPhase` added alongside, still valid YAML |
| Externals group renamed | exits 1, no leftover `.tmp` |
| `xcodegen` absent | exits 1 with the `brew install xcodegen` hint |

The end-to-end proof of the original fix is the archive: the 19 MB ipa passed Apple's validation, the
56 MB one did not.

### DIFF
--- a/scripts/build-ios.sh
+++ b/scripts/build-ios.sh
@@ -57,16 +57,61 @@ rustup target add aarch64-apple-ios
 echo "[2/4] Installing npm dependencies..."
 npm install
 
-if [ ! -d "gen/apple" ] || [ "${FORCE_INIT:-0}" = "1" ]; then
+GEN_APPLE="src-tauri/gen/apple"
+
+# `tauri ios init` lists Externals/ as a plain source group, so XcodeGen puts it in BOTH "Link
+# Binary With Libraries" and "Copy Bundle Resources". The second one copies the Rust static library
+# into the app bundle, which App Store validation rejects outright:
+#
+#   Invalid bundle structure. The "Kite Ground Control.app/libapp.a" binary file is not permitted.
+#
+# It also carried the library's ~37 MB into the ipa for nothing. libapp.a is linked, never a
+# resource, so the group needs `buildPhase: none`. src-tauri/gen/apple is generated and gitignored,
+# so this cannot be fixed by committing the project file: it has to be re-applied on every build.
+patch_externals_build_phase() {
+    local yml="$GEN_APPLE/project.yml"
+    if [ ! -f "$yml" ]; then
+        echo "[ERROR] $yml not found. Run with FORCE_INIT=1 to regenerate the Xcode project."
+        exit 1
+    fi
+    # Already carries a buildPhase? Leave the file alone.
+    local has_phase='$0 ~ /^[[:space:]]*- path: Externals$/ { getline nxt; if (nxt ~ /^[[:space:]]*buildPhase:/) found = 1 } END { exit !found }'
+    if awk "$has_phase" "$yml"; then
+        echo "      Externals already excluded from Copy Bundle Resources."
+        return 0
+    fi
+    awk '
+        { print }
+        /^[[:space:]]*- path: Externals$/ {
+            match($0, /^[[:space:]]*/)
+            indent = substr($0, 1, RLENGTH)
+            print indent "  buildPhase: none"
+        }
+    ' "$yml" > "$yml.tmp"
+    # Never build on a silent miss: a template change in Tauri that renames or re-indents the group
+    # would otherwise put libapp.a back in the bundle and the rejection would only surface at upload.
+    if ! awk "$has_phase" "$yml.tmp"; then
+        rm -f "$yml.tmp"
+        echo "[ERROR] Could not exclude Externals from Copy Bundle Resources in $yml."
+        echo "        Add 'buildPhase: none' under the Externals source group by hand, or the ipa"
+        echo "        will contain libapp.a and App Store validation will reject it."
+        exit 1
+    fi
+    mv "$yml.tmp" "$yml"
+    echo "      Excluded Externals from Copy Bundle Resources (keeps libapp.a out of the app)."
+}
+
+if [ ! -d "$GEN_APPLE" ] || [ "${FORCE_INIT:-0}" = "1" ]; then
     echo "[3/4] Generating the Xcode project (tauri ios init)..."
     npm run tauri ios init
 else
-    echo "[3/4] gen/apple already present — skipping init (FORCE_INIT=1 to redo)."
+    echo "[3/4] $GEN_APPLE already present — skipping init (FORCE_INIT=1 to redo)."
 fi
+patch_externals_build_phase
 
 echo "[4/4] Building signed release .ipa with Tauri..."
 npm run tauri ios build
 
 echo ""
 echo "[build-ios] Done. Look for the .ipa under:"
-echo "  gen/apple/build/arm64/  (and the Xcode archive/export path Tauri prints above)"
+echo "  $GEN_APPLE/build/arm64/  (and the Xcode archive/export path Tauri prints above)"
--- a/scripts/build-ios.sh
+++ b/scripts/build-ios.sh
@@ -57,57 +57,12 @@ rustup target add aarch64-apple-ios
 echo "[2/4] Installing npm dependencies..."
 npm install
 
-GEN_APPLE="src-tauri/gen/apple"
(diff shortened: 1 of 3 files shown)


b14ckyy (changes requested):
Thanks, the diagnosis is right and the path fix is real. I checked the mechanism against the sources: XcodeGen has no default build phase for `.a` and falls back to `resources` (`SourceGenerator.getDefaultBuildPhase`), so the library goes into "Copy Bundle Resources"; the link itself comes from the `dependencies: - framework: libapp.a` entry, not from the source group. Upstream noticed the same thing on 2026-08-31 (tauri-apps/tauri#15890 adds `excludes: ["**/*.a"]` to the Externals group), but it is not in any released CLI yet (2.11.4 still has the bare entry). Your patch is compatible with that future template: I ran `patch_externals_build_phase` against the 2.10.1 form, the dev form with `excludes`, an already patched file and a renamed group. All four behave as in your table, no `.tmp` is left behind, and the result parses as YAML with both keys on the group.

One real problem though: **the patched `project.yml` never reaches the Xcode project.**

Only `tauri ios init` runs xcodegen (`crates/tauri-cli/src/mobile/ios/project.rs` is the single call site in the whole CLI). `tauri ios build` loads the existing pbxproj and synchronizes values into it (`ensure_init`, `synchronize_project_config`), it does not regenerate it. In the script the patch is applied *after* init, and init only runs when `src-tauri/gen/apple` is missing, so nothing consumes the change. The 19 MB ipa is clean because a fresh init runs with an empty `Externals/`, and the path fix stops the re-inits that used to pick the built library up. The patch has no part in it.

Where it breaks: anyone who has used `scripts/run-ipad.sh` before has a pbxproj generated with a populated `Externals/`, i.e. with `libapp.a` in Copy Bundle Resources. This script then skips init, patches the yml to no effect, builds with the stale pbxproj, and the archive is rejected again.

Two changes needed:

1. Regenerate after patching, with the same call Tauri uses at init:
   ```bash
   xcodegen generate --spec "$GEN_APPLE/project.yml"
   ```
   A regenerated pbxproj loses nothing a re-init would not lose as well, and `tauri ios build` writes team and config back into it anyway.
2. Give `scripts/run-ipad.sh` the same treatment (path fix + patch as a shared helper). It still re-inits on every run and, with an unpatched yml, produces the broken pbxproj again. Upstream's comment on #15890 also names the failure mode once several copies of `libapp.a` exist side by side ("Multiple commands produce …/libapp.a"), which is exactly what a debug + release or device + simulator mix in `Externals/` produces.

Side note on the description: "a hand-edit of `project.yml` never survived" does not match the CLI. `gen()` in `mobile/ios/project.rs` only creates files that do not exist yet (`if !path.exists() { create } else { Ok(None) }`) and then runs xcodegen, so an edited yml is kept and honoured on the next init. Harmless, but worth correcting in the text.

Please also add the `bug` label and the 1.1 milestone.

SebastianKumor:
You were right, and I measured it rather than taking it on trust: with a marker setting in `project.yml`, a full `tauri ios build` reached `xcodebuild` without the marker ever landing in `project.pbxproj` (mtime unchanged). My patch was doing nothing; the clean 19 MB ipa came from the fresh init plus the path fix, exactly as you said.

Fixed: `xcodegen generate` after patching (only when the patch changed something or the pbxproj still copies the library), plumbing moved to `scripts/lib/ios-project.sh`, and `run-ipad.sh` uses it with its own path fix. Signing team survives, it lives in the yml.

Six cases tested incl. your dev-run failure state and the `excludes` form; table is in the updated description, along with the corrected hand-edit claim.

I cannot set labels or milestones on this repo (no write access), so the `bug` label and 1.1 milestone need you.

b14ckyy (approved):
Merge-ready. Both changes from the review are in and I checked them against the sources again:

- `ios_regenerate_project` makes the same call `tauri ios init` makes (`xcodegen generate --spec <yml>`, cargo-mobile2 `apple/project.rs`), and XcodeGen generates next to the spec by default, so running it from the repo root is fine. The regeneration is gated on "the patch changed the file or the pbxproj still copies the library", and the post-check greps `libapp.a in Resources` in `kite-gc.xcodeproj/project.pbxproj`, which is the name cargo-mobile2 derives from the crate name.
- `scripts/run-ipad.sh` now shares `scripts/lib/ios-project.sh`, path fix included, so a dev run no longer leaves the pbxproj a later archive is rejected for.
- The description matches the CLI now.

One nit, not blocking: init passes `--no-env` to xcodegen, the regeneration does not. XcodeGen substitutes `${NAME}` from the shell environment, and the template carries `${FORCE_COLOR}` and `${CONFIGURATION}` bare, so a shell that exports one of them bakes the value into the regenerated project. Adding `--no-env` keeps the two paths identical; a one-token follow-up whenever you touch the file next.

Thanks for measuring the build path rather than taking it on trust.

b14ckyy (approved):
Merge-ready. Both changes from the review are in and I checked them against the sources again:

- `ios_regenerate_project` makes the same call `tauri ios init` makes (`xcodegen generate --spec <yml>`, cargo-mobile2 `apple/project.rs`), and XcodeGen generates next to the spec by default, so running it from the repo root is fine. The regeneration is gated on "the patch changed the file or the pbxproj still copies the library", and the post-check greps `libapp.a in Resources` in `kite-gc.xcodeproj/project.pbxproj`, which is the name cargo-mobile2 derives from the crate name.
- `scripts/run-ipad.sh` now shares `scripts/lib/ios-project.sh`, path fix included, so a dev run no longer leaves the pbxproj a later archive is rejected for.
- The description matches the CLI now.

One nit, not blocking: init passes `--no-env` to xcodegen, the regeneration does not. XcodeGen substitutes `${NAME}` from the shell environment, and the template carries `${FORCE_COLOR}` and `${CONFIGURATION}` bare, so a shell that exports one of them bakes the value into the regenerated project. Adding `--no-env` keeps the two paths identical; a one-token follow-up whenever you touch the file next.

Thanks for measuring the build path rather than taking it on trust.

PR #151, opened 2026-09-10
feat(connection): default-protocol setting and a protocol-aware network port

SebastianKumor:
## Description

The two follow-ups from your review of #131.

**Default protocol.** `Settings -> Connection -> Default Protocol`, built the way you described it:
a dropdown on **Last used** by default, which is exactly what Kite does today, and settable to MSP,
MAVLink or Telemetry, in which case that choice overrides the stored last-used protocol on every
launch. `lastProtocol` keeps being written whatever the setting says, so switching back to
**Last used** resumes where the pilot left off rather than starting blank.

The startup restore moves into a `resolveStartupProtocol()` helper. That is deliberate: its body
also contains the `telemetry` restore fix from #143, so when `release/1.0.x` forward-ports there is
one obvious hunk to resolve in favour of `development` instead of a conflict spread over the
surrounding lines. Nothing else about that fix changes here.

**The network port follows the protocol**, not just the transport: MAVLink UDP 14550 or TCP 5760,
MSP TCP 5761. Before, selecting MAVLink over TCP still offered INAV SITL's MSP port and the pilot
had to know the right number. A port typed by hand is untouched by both selectors, and Telemetry is
passive so it keeps whatever is set. This only brings the port in line with the baud rate, which
already presets 115200 for MSP and 57600 for MAVLink.

Both compose with your per-protocol hide checkboxes when you get to them: a hidden protocol drops
out of the dropdown, and the port map is keyed by protocol already.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] No unrelated changes

## Notes for reviewer

`npm run check` 582 files / 0 errors / 0 warnings, `npm run build` clean, `cargo check` clean (no
Rust in this PR). **The one box I left unticked is the honest one:** I traced the whole path
statically and built it, but I have not sat in front of the running app and worked the dropdown
through the four options and a restart. Tell me if you want that before merging and I will run it
against the simulator and report back.

Points worth your attention:

1. **Existing profiles.** `defaultProtocol` is absent from stored settings, and the loader spreads
   `defaults` first, so they all come up as `last`: no behaviour change for anyone who never opens
   the setting.
2. **`5760` is now treated as a default** rather than a custom port, so it is re-defaulted when the
   protocol changes. It is ArduPilot SITL's TCP port, so I think that is right, but it is a
   judgement call and easy to drop from the map.
3. **The MSP-over-UDP entry** is 14550 only because there is no standard MSP UDP port, so it needs
   some value rather than a blank. Documented as a starting point in the connecting guide.
4. Docs touched: the settings reference (new Connection section), the port table and Reconnecting
   section in the connecting guide, and the network bullet in connection troubleshooting.
5. Strings are in all five locales (en, de, fr, bg, zh). The pre-existing gaps in fr/bg/zh are
   untouched.


SebastianKumor:
Pushed two fixes on top after a review pass found them, both in the port logic:

- Deciding "did Kite fill this in" from the number was wrong: 5760 is ArduPilot SITL's port and a fine choice for an MSP bridge, so a typed 5760 got rewritten on the next switch. Now tracked in `portIsAuto`, persisted, owned by `+page` so the toolbar and phone popout share one answer.
- Startup had the same gap: a pinned MAVLink after an MSP/TCP session came up on 5761. The port now follows the resolved protocol, unless the pilot typed it.

Checked 10 cases incl. typed-5760 surviving a switch and a restart. `npm run check` 582/0, build clean.

b14ckyy (approved):
Merge-ready. This is the setting exactly as we discussed on #131: "Last used" as the default with the historic behaviour, a pinned value winning over the store on every launch, `lastProtocol` still written on every connect so switching back resumes where the pilot left off, and existing profiles coming up as "Last used" because the loader spreads the defaults first. Keeping the #143 restore fix intact inside `resolveStartupProtocol()` so the `release/1.0.x` forward-port meets one obvious hunk is a good call. The port flag living in `+page` so the toolbar and the phone popout cannot disagree, the startup move of the port when a pinned protocol overrides the stored one, the five locales, the changelog placement (Added / Improved) and the three docs pages all check out. I will run it through the simulator myself before merging, so no need for you to do the runtime pass.

Two observations on the port flag, neither blocking:

1. Once a port has been typed there is no way back to "Kite's": the flag is persisted and no selector ever moves the port again, even if the pilot later types exactly 14550. Before, the number itself was the way back. A one-liner would restore that: when the typed value equals the standard port for the current selection, treat it as Kite's again.
2. Existing MAVLink-over-TCP profiles sit on 5761 (the old transport-only default). The legacy inference compares against the new 5760 and classifies them as typed, which is the right choice, since it never moves a working endpoint. The price is that those profiles lose the re-defaulting for good, which point 1 would give back.

Two notes from my side, for your feedback rather than for this PR:

- I have plans for the port numbers anyway: either a dropdown of standard ports, or a list of the last five ports used per transport. I will do that later, unless you feel like picking it up.
- 5760 is the SITL's own port, but in practice Mission Planner grabs it, so I have only ever reached SITL over 5761 or 5762. That is exactly what led me to the standard-port dropdown idea. Fine as the MAVLink/TCP default here, just so the reasoning is on record.

SebastianKumor:
Took your observation 1 as a separate commit (`6d84aa3`), so you can drop it and keep the diff you approved if you prefer: typing the standard port for the current selection marks the port as Kite's again, anything else keeps it the pilot's. That gives observation 2 back too, since a legacy MAVLink/TCP profile on 5761 can type 5760 and get re-defaulting again.

Matrix is 17 cases now: a partial `1455` is still the pilot's, and a typed 5760 on MSP/TCP stays the pilot's since 5761 is the standard there.

On the port numbers: the dropdown of standard ports plus last-five-per-transport is yours, I would rather not cut across it. Good to know about Mission Planner holding 5760, that is a better argument for the dropdown than anything in this PR.

PR #182, opened 2026-09-13
test(relay): bind the TCP sink test on port 0 instead of a probed port

b14ckyy:
Flake seen on the macOS job of #181: `telemetry_forward::output::tcp::tests::a_client_that_never_reads_is_dropped_instead_of_blocking_the_sink` — "TCP relay bind 0.0.0.0:49232 failed: Address already in use".

The test probed a free port (bind `127.0.0.1:0`, read the port, drop the probe) and had the sink bind it a moment later. On macOS the ephemeral range is 49152–65535 and the tests run in parallel, so another socket can take the port in that window.

`TcpSink` now records the bound address and port (`port()` accessor); the test opens port 0 and asks the sink which port it got — no probe-and-release window. Side effect: the relay status line shows the real bound port. No behaviour change for a configured port.

`cargo test --lib telemetry_forward` green, `cargo check` and clippy clean on the file.

Affects: v1.0.0 and 1.0.1 (test only). Merge-up into `development` with the pending forward-port.


b14ckyy:
Test-only change — it does not belong on the release line (real bugs and docs only). Re-opened against development.

Issue #136, opened 2026-09-08
Multi-vehicle support (connect and control several aircraft at once)


dldnjsxo95:
Kite currently handles one vehicle link at a time (single `ActiveProtocol` slot, MAVLink frames from other sysids are dropped). It would be great to connect to several aircraft at once, see all of them on the map, and switch the active vehicle for HUD / control / mission upload without reconnecting – like the vehicle selector in QGroundControl.

b14ckyy:
Hey, thanks for your request. 
I can have a look at that in the future. But since this is a very dedicated feature that is probably barely needed in the hobby community, it won't have priority. It would also need external help since I have no fleet of Mavlink Capable UAV at hand that I can do actual testing with. Additionally this would be a significant change in the internal Telemetry and Database Pipeline that needs very thorough testing. Hence I will keep it in mind for the future.

b14ckyy:
Some groundwork so this ticket has a concrete reference. It is a research summary, not a commitment — the feature stays unscheduled for the reasons in the comment above, but the idea is parked, not rejected.

**What "multi-vehicle" means, in three sizes**

| Level | User-visible result | Cost |
|---|---|---|
| A — Observe | Other MAVLink systems on the *same* link show up on the map (position, heading, sysid, armed). No HUD, no control. | Small. Kite's Radar subsystem already has a reserved slot for radio-sourced contacts; this is a tap on the MAVLink receive path. |
| B — Switch | Several vehicles connected (one shared link or one link per vehicle), one *active* vehicle owning HUD / control / mission / params / recording, the rest drawn on the map, switching without reconnecting. This is what the ticket asks for. | Large. See below. |
| C — Fleet | Commands to all/selected vehicles, per-vehicle alerts, stream-rate budgeting on a shared radio. | Medium on top of B. |

**Why B is a big change in Kite** (audit of the current code, `development` as of today)

- The app has exactly one protocol slot and one handler thread per link; the MAVLink receive path locks onto the first autopilot heartbeat and silently drops every other system id. Mission / param / command replies are correlated by message type, not by source sysid, so two vehicles cannot even transfer a mission at the same time.
- The ~30 internal telemetry events carry no vehicle key, and four independent consumers (frontend store, relay/forwarding hub, Telemetry API, link status) assume a single producer.
- The frontend telemetry store is a singleton read by 42 files, including every widget and both maps.
[...]

PR #194, opened 2026-09-15
fix(transport): UDP — stable fallback port, fan-out to every live peer, peer loss is not link loss

dldnjsxo95:
## Description

Four changes to the UDP transport, one small UI fix that the UDP work made visible, and — from the review — a MAVLink stall watchdog so the status bar tells the truth while a link is stalled. Split out of the multi-vehicle branch (#136) as requested; everything here is reachable with a **single vehicle**, so it changes single-vehicle behaviour and deserves its own review.

### Intended behaviour

| Situation | Before | After |
|---|---|---|
| Local port (e.g. 14550) is free | Bind it, listen there | Unchanged |
| Local port is taken (another GCS / Mission Planner listening) | Bind an **ephemeral** port; the log said so, the UI showed a bare "no HEARTBEAT" | Bind a **stable** alternate first (`target + 10000`, so 14550 → 24550), ephemeral only if that is busy too. A failed handshake carries a note naming the busy port and what to do about it. |
| Who receives our outgoing frames | The **last** source we heard from (single peer, re-targeted on every datagram) | **Every** source heard within the last **10 s**. Silent peers age out; a send failure to one of several peers is ignored (it ages out). With no peer learned yet, the configured `host:port` gets the frame — unchanged, this is how the first GCS HEARTBEAT wakes a client-mode FC or bridge. |
| A peer closes its socket (Windows: ICMP port-unreachable → `recv_from` returns `ConnectionReset`) | Mapped to `TransportError::Disconnected` → MAVLink handler exited → link dead (and the UI didn't notice, see below) | Treated like a read timeout (`Ok(0)`); the peer ages out of the fan-out. Also covers `ConnectionRefused` / `ConnectionAborted`. |
| Our own first HEARTBEAT comes back from 127.0.0.1:14550 (the default target is our own port) | Learned as a peer; with the fan-out every later frame also went to ourselves and was parsed by our own handler | Datagrams from our own loopback port are dropped before peer learning (`fix(transport): UDP — never learn our own socket as a peer`). |
| The FC goes quiet on an open transport (radio dropout, SITL paused, UDP peer gone) | Nothing: the MAVLink handler had no liveness signal, so the status bar stayed "connected" with frozen values (MSP links showed "Reconnecting…" via the scheduler's stall watchdog) | After **3 s** without any FC-origin frame: `telemetry-fc-link { alive: false }` once + one `warn` at the default level; on the next FC frame `{ alive: true }` + an `info`. **No teardown, no timeout, no prompt** — the handler runs on unchanged; `StatusBar.svelte` already renders the event. (`feat(mavlink): stall watchdog`, from the review.) |
| Backend emits `mavlink-disconnected` (transport died) | No frontend listener — UI stayed on "connected" until the user disconnected by hand | Handled exactly like `connection-lost`: connection state torn down, reconnect-on-return flag set when hidden. |

Why the stable port: relays such as Mission Planner's UDP mirror remember the client address they stream to and keep streaming to a dead port for a while after we close it. Coming back on the same port lets a reconnect pick the stream up immediately instead of waiting for the relay to forget us.

Why the fan-out: several sources pushing to our listening port each speak from their own socket (SITL instances, a telemetry bridge next to a GCS mirror). A command sent only to "whoever spoke last" mostly landed on the wrong one and was ignored. Vehicles drop frames not addressed to them, so sending to all recent peers is harmless. With one peer (the normal single-vehicle case) the send set has one member and behaviour is identical to before.

### What to look at

- `stable_fallback_port()` — `target + 10000`, wraps down for targets above 55535.
- `write_bytes()` — with **one** peer a send error is still an error; with several, one failing peer is not.
- `read_bytes()` — the three "connection" error kinds now return `Ok(0)` instead of ending the link. On Linux/macOS these kinds do not normally occur for UDP, so the change is Windows-relevant in practice.
- The `diagnostic_note()` trait method defaults to `None`; only UDP implements it. `connect()` reads it once before the transport moves into the protocol path and appends it to a handshake error.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] `cargo check` (in `src-tauri`) passes (clippy clean in the touched files, `cargo test` green)
- [x] I have tested the changes locally (dev, Windows 11 25H2, PX4 v1.18 SITL `gz_x500` in WSL2 with mirrored networking, MAVLink to 127.0.0.1:14550):
  - free port: bind on 14550, only the SITL's source port 18570 learned as a peer (our own echo no longer shows up), handshake in 0.4 s;
  - 14550 held by another process: bound **24550**, handshake timed out after 10 s and the error carried the note ("Local UDP port 14550 is already in use … Kite listened on 24550 instead …");
  - SITL killed while connected: `recv_from` returned `ConnectionReset` (WSAECONNRESET) once per outgoing frame — logged at debug as "a peer went away, ignoring", link and UI stayed up. With the stall watchdog: 3 s after the kill the log shows one `[WARN] Link stalled — no MAVLink frame from the FC for 3s (transport still open)` and `fcLinkAlive` flipped to `false` while `connection.status` stayed `connected`; on the SITL restart (~15 s later) one `[INFO] Link recovered — MAVLink frames from the FC resumed` and `fcLinkAlive` back to `true` — no disconnect, no reconnect;
  - `mavlink-disconnected`: connected over TCP to a scripted MAVLink v1 heartbeat source that closes its socket after 10 s → backend "MAVLink transport disconnected" → frontend tore the connection down, status `disconnected`.
  - Not exercised: a Mission Planner UDP mirror (that is the stable-port scenario in the wild) and several vehicles on one port with a single-vehicle build (fan-out has one member there).
- [ ] Documentation (ROADMAP / CHANGELOG) updated if needed — changelog line suggested below
- [x] No unrelated changes

## Notes for reviewer

- **Bug-fix PR:** the dropped link on peer loss (`ConnectionReset`) and the stale "connected" UI exist in the **released 1.0 version**; the port fallback / fan-out are behaviour improvements on top. Affects: v1.0.x, all firmwares, UDP (and for the UI fix any MAVLink) links.
- Four commits, one topic each: the transport (`fix(transport)`, two commits — the second drops our own echo), the UI listener (`fix(ui)`), and the stall watchdog (`feat(mavlink)`, from the review). Happy to move the UI one into its own PR if you prefer.
- Suggested test on your side: Mission Planner listening on 14550 with a UDP mirror to Kite (exercises the stable fallback + the note), plus SITL pushing to 14550 while a second GCS on the same port comes and goes (exercises fan-out and peer loss).
- Suggested changelog line: "UDP links: a busy local port falls back to a stable alternate and is reported on a failed handshake; a peer going away no longer ends the link; commands reach every recent sender."


b14ckyy (changes requested):
Read the whole diff against the code — the transport part is right, and it is the design we want: a peer that goes quiet is a **stall**, not a lost link. Kite keeps the session, keeps polling / sending heartbeats, and picks the vehicle up again the moment it is back, with nothing for the user to do (ADR-042: only a *physical* transport loss — USB unplugged, BLE gone — may tear a connection down; a telemetry stall never does). The Windows `ConnectionReset` teardown was a genuine 1.0 bug against that rule, so thanks for catching it; the stable fallback port, the fan-out, the own-echo drop and the `mavlink-disconnected` listener (a real 1.0 bug too — the event had no listener since day one) all read correct.

**One change before merge**, and it is about honesty, not about disconnecting:

> "link and UI stayed up (heartbeat loss takes over from there)"

There is nothing to take over. The MAVLink handler has **no liveness signal at all** — `telemetry-fc-link` (the "Reconnecting…" state of the status bar) is only emitted by the MSP scheduler (`scheduler/mod.rs`, stall watchdog, `STALL_WARN_AFTER` = 3 s, `FcLinkAlive { alive }`) and by the passive-telemetry path. On an MSP link a radio dropout therefore shows "Reconnecting…" and goes back to "connected" by itself; on a MAVLink link the user sees "connected" with frozen values and cannot tell Kite from the radio. With the reset now swallowed that gap is the same on every platform, so this PR is the place to close it:

- in `mavlink_proto/handler.rs`, remember the time of the last frame received from the FC's sysid;
- after 3 s of silence (the MSP value; count any FC-origin frame, not just HEARTBEAT) emit `telemetry-fc-link { alive: false }` once and `log::warn!` once at the default level ("Link stalled — no MAVLink frame for Ns (transport still open)"), like the MSP watchdog does;
- on the next FC frame emit `{ alive: true }` (a `log::info!` is enough there).

To be explicit: **no teardown, no timeout-disconnect, no prompt** — the handler keeps running exactly as it does now; the event only drives the status bar (`StatusBar.svelte` already renders it for MSP, nothing to add on the frontend). That is the counterpart of "peer loss is not link loss": the link stays, and the UI says so truthfully.

Two notes, no action needed:
- Fan-out corner case worth a sentence in the user docs later: Kite connected to a vehicle directly *and* fed by another ground station's UDP mirror on the same port means every command goes out twice (the mirror forwards it). One peer — the normal case — is unchanged.
- I'll test the Mission Planner mirror (stable port + the note) and a SITL fan-in with several instances on my side before merging.

dldnjsxo95:
Done — and thank you for the correction; "heartbeat loss takes over" was wrong, there was nothing there. Pushed as one more commit, `feat(mavlink): stall watchdog — tell the status bar when the FC goes quiet on an open link`:

- `mavlink_proto/handler.rs` remembers the time of the last frame from the FC's sysid (any FC-origin frame, not just HEARTBEAT).
- After 3 s of silence (MSP's `STALL_WARN_AFTER`): `telemetry-fc-link { alive: false }` once and one `log::warn!` — `Link stalled — no MAVLink frame from the FC for 3s (transport still open)`.
- On the next FC frame: `{ alive: true }` and one `log::info!` (`Link recovered — MAVLink frames from the FC resumed`).
- No teardown, no timeout-disconnect, no prompt; the payload struct is module-local like the other event payloads in the handler, nothing on the frontend.

Verified against PX4 SITL over UDP: kill the SITL → 3 s later the warning and `fcLinkAlive = false` while `connection.status` stays `connected`; restart the SITL → the info line and `fcLinkAlive = true`, same session, no reconnect. The PR text is updated (the table has a row for it, the test list carries the run). Your two notes are noted — the fan-out corner case (direct link + a mirror on the same port → commands go out twice) goes into the user docs with the feature.

b14ckyy (approved):
Read the watchdog commit against the loop: the check sits after the read step and the 50 ms read timeout falls through to it, so the 3 s threshold is reached without a single frame, and only FC-sysid frames reset it. Exactly the counterpart we asked for, thanks. `cargo test` and `svelte-check` pass locally on the head. Approving; the Mission Planner mirror and a SITL fan-in run on my side before the merge.

b14ckyy:
One small thing surfaced while re-reading the fan-out (`udp.rs`, `write_bytes`), not blocking: with two or more peers the function returns `Ok` even when every `send_to` failed — the match only returns the error for the single-peer case, while the comment says "at least one peer took it". Counting the failures and returning the first error when all of them failed makes the comment true. Two lines, no behaviour change for the normal one-peer case — either here or as a follow-up, your call.

PR #195, opened 2026-09-15
fix(px4): mission, parameter and take-off/land fixes for PX4 (+ autopilot component addressing)

dldnjsxo95:
## Description

The PX4 fixes from the multi-vehicle branch (#136), split out as requested. Nine commits (seven + two from the review), one topic each; every commit compiles on its own and carries an `Affects:` line. Found while flying PX4 SITL (v1.18-dev) with plans written in Mission Planner and QGC; the ArduPilot-relevant items are marked because that path is what 1.0 users fly today.

### Commits and what changes on the wire

| # | Commit | What | Affects |
|---|---|---|---|
| 1 | `fix(mavlink): address the autopilot component in the mission microprotocol` | Every mission/fence/rally message targets component **1** (`MAV_COMP_ID_AUTOPILOT1`) instead of **0**. COMMAND_LONG/INT already did. | **ArduPilot + PX4.** Direct links: no visible change (both accept either). Through a routing relay (Mission Planner's MAVLink mirror) uploads no longer stall. |
| 2 | `fix(mission): PX4 NaN mission params — normalise on download, tolerate null on upload` | Non-finite params/alt from the FC become 0 ("not set"); JSON `null` in an incoming plan reads as 0; the frontend serializer/parser never throw on NaN. | PX4 (the NaN source). Firmware-neutral hardening; ArduPilot never sends NaN here. |
| 3 | `fix(mission): PX4 upload — unset yaw goes out as NaN, action items in MAV_FRAME_MISSION` | (a) On PX4, param4 = 0 on yaw-bearing position items is sent as NaN so the vehicle's yaw mode applies (QGC does the same). **Gated on PX4.** (b) ~~frame forced in `wp_to_item()`~~ — **superseded by commit 8** (see below). | (a) PX4 only. |
| 4 | `fix(params): PX4 integer parameters — decode byte-cast PARAM_VALUE, send typed PARAM_SET` | Integer-typed params from PX4 are byte-cast floats; decoded by declared type. PARAM_SET on PX4 reads the type first and writes it byte-cast with that type. Unit tests included. | PX4 only. ArduPilot: values that are plain numbers pass through unchanged; PARAM_SET stays REAL32 as before. |
| 5 | `fix(fence): ask each firmware only for its own fence / rally parameters` | FENCE_* for ArduPilot, GF_* for PX4; no RALLY_* on PX4. | PX4 (fewer `Unknown param name` timeouts on connect). ArduPilot: no longer asks for the three GF_* names it never had. |
| 6 | `fix(control): take-off / land with NaN yaw and position — "here", not 0°N 0°E` | `NAV_TAKEOFF` / `NAV_LAND` send NaN for yaw/lat/lon instead of 0. | **ArduPilot + PX4.** PX4: the fix (it climbed towards 0°N 0°E). ArduPilot ignores these fields for take-off (param7 only) and land-in-place, so NaN is harmless — but it is on the wire. |
| 7 | `feat(rc): PX4 — show the live COM_RC_IN_MODE and offer the one-click fix` | RC panel reads COM_RC_IN_MODE on a PX4 connect, warns when it blocks joystick input (0 / 4), offers "set 2" and a re-read. Four new `en.json` keys. | PX4 only (UI shown only for PX4). |
| 8 | `fix(mission): decide the action-item frame in the planner, not in the upload path` (review) | `wp_to_item()` passes the frame through unchanged again, so fence / rally items keep the frames their editors set. `framesForUpload()` in `missionArdupilot.ts` sets `MAV_FRAME_MISSION` on every command the catalog knows that is **not** in ArduPilot's `stored_in_location()` set (listed verbatim: 16-19, 20 kept as-is, 21, 22, 30, 31, 82, 84, 85, 92, 94, 179, 188, 189, 191, 195, 201); unknown commands keep their frame. Runs only on the MISSION upload in `ArduMissionPanel`. | **ArduPilot + PX4** for mission uploads (action items now carry frame 2 on the wire); fence / rally uploads back to 1.0 behaviour. |
| 9 | `i18n(de): COM_RC_IN_MODE banner strings` (review) | Your four German strings. | — |

### Items to re-run against ArduPilot SITL (your request)

- **Commit 1** — `target_component` 0 → 1 on MISSION_REQUEST_LIST / REQUEST_INT / COUNT / ITEM_INT / CLEAR_ALL / ACK. Mission, fence and rally download/upload/clear.
- **Commit 8** (replaces 3(b)) — action items uploaded in `MAV_FRAME_MISSION`, decided in the frontend from the catalog. A plan with DO_CHANGE_SPEED / DO_SET_ROI / CONDITION_DELAY etc.; ArduPilot stores them without a frame, so a download→upload round trip should be identical. Plus a **fence and a rally write** ("Save to FC" in both editors) — the path 3(b) broke.
- **Commit 6** — Take-off / Land with NaN in param4–6.

Everything else is gated on the PX4 `fc_variant` or only reachable on PX4.

## Type of change

- [x] Bug fix
- [x] New feature (commit 7 only)
- [ ] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes
- [x] `cargo check` (in `src-tauri`) passes — clippy clean in the touched files, `cargo test` green (164, incl. 4 new)
- [x] I have tested the changes locally (dev, Windows 11 25H2, PX4 v1.18.0-beta SITL `gz_x500` in WSL2, MAVLink/UDP 127.0.0.1:14550):
  - **params (4, 7):** `COM_RC_IN_MODE` read 1 (INT32, byte-cast on the wire), set 2 → read back 2, restored 1; `SYS_AUTOSTART` read 4001; `MPC_XY_VEL_MAX` (REAL32) read 12.
  - **fence / rally (5):** `fence_read_all` returned `GF_ACTION / GF_MAX_HOR_DIST / GF_MAX_VER_DIST` only, `rally_read_all` no params; the PX4 console logged **no** `Unknown param name` during either read (before this change it did for every FENCE_* / RALLY_* name).
  - **mission (1, 2, 3):** uploaded TAKEOFF, WAYPOINT (yaw 0 = unset), DO_CHANGE_SPEED (frame 0 as Mission Planner writes it), WAYPOINT (yaw 90), LAND — accepted (no `Invalid item`), downloaded 5/5: the DO item came back in frame 2 (`MAV_FRAME_MISSION`), the unset yaw as 0 (NaN normalised), the explicit yaw as 90. With `NAV_RETURN_TO_LAUNCH` in the list PX4 rejects the item — PX4 has no RTL mission item, that is not a Kite bug.
  - **take-off / land (6):** arm → take-off 8 m → climbed 7.6 m with **0.1 m** horizontal drift over 12 s (previously it headed for 0°N 0°E) → land → "Landing at current position", disarmed 0.1 m from the start.
  - **after the review (8):** mission upload through `framesForUpload` — frames sent `[3, 3, 2, 3, 3]` for TAKEOFF / WP / DO_CHANGE_SPEED (file frame 0) / WP / LAND, accepted, downloaded 5/5 with the DO item in frame 2; **fence write round trip** — one inclusion polygon (4 vertices) + `GF_ACTION = 1` written, read back identical (vertices, kind, shape, param), then cleared (0 zones, `GF_ACTION` restored to 2); **rally write round trip** — one point (lat/lon/alt_cm) written, read back identical, cleared. No warning on the PX4 console during any of it.
  - Not run: ArduPilot SITL (commits 1, 6, 8 on the ArduPilot side — your list above); a real vehicle.
- [ ] Documentation (ROADMAP / CHANGELOG) updated if needed — changelog lines suggested below
- [x] No unrelated changes

## Notes for reviewer

- Bug-fix PR: commits 1, 6 and 8 change what the **released 1.0 version** sends to ArduPilot; the rest is PX4-only or gated. Commits 2 and 3 also fix a UI freeze that exists in 1.0 for anyone downloading a PX4 mission.
- `de.json` carries your four strings (commit 9); the other locales still fall back to English.
- Suggested changelog lines: "PX4: take-off and land no longer aim at 0°N 0°E; missions with unset yaw or action items upload; integer parameters read and write correctly; the RC panel shows COM_RC_IN_MODE and can set it." / "MAVLink: mission transfers address the autopilot component, so they pass through routing relays (Mission Planner mirror)."


b14ckyy (changes requested):
Thanks — this is a clean split, and the `Affects:` lines made checking the ArduPilot side easy. I verified commit 6 against current ArduPilot master (`convert_COMMAND_LONG_loc_param()` maps NaN to 0, `NAV_TAKEOFF` deliberately ignores lat/lng, Copter ignores param4), so that one is safe; commits 1, 2, 4, 5 and 7 read correct.

**Commit 3(b) needs a change — it breaks the geofence and rally editors on ArduPilot.** `wp_to_item()` now forces `MAV_FRAME_MISSION` for every command not in `cmd_has_location()`, but the same function serves `fence_write_all` / `rally_write_all`, whose items (5001–5004, 5006, 5100) are not in that list. ArduPilot rejects them: `MissionItemProtocol_Fence.cpp` (frame must be GLOBAL / RELATIVE / TERRAIN) and `MissionItemProtocol_Rally.cpp` (same switch) both return `MAV_MISSION_UNSUPPORTED_FRAME`. "Save to FC" in both editors fails — the read paths you tested don't hit it.

The list itself has a second problem: `2500 | 2501` are `VIDEO_START/STOP_CAPTURE`; `NAV_PAYLOAD_PLACE` is **94**. A payload-place item therefore goes out in `MAV_FRAME_MISSION`, and ArduPilot does *not* reject that for a location command — `AP_Mission::mavlink_int_to_mission_cmd` treats `MAV_FRAME_MISSION` like `MAV_FRAME_GLOBAL`, i.e. **absolute altitude**. 30 m relative silently becomes 30 m AMSL. Also missing versus ArduPilot's `stored_in_location()`: 92 `NAV_GUIDED_ENABLE`, 188 `DO_RETURN_PATH_START`, 191 `DO_GO_AROUND`.

Suggested fix: don't decide the frame in the backend at all. The frontend already owns the command catalog (`arduCommandCatalog.ts` — `cmdHasLocation()` / `cmdStandaloneCoordinate()`); let the mission upload path set frame 2 on pure action items before it invokes, and leave `wp_to_item()` passing the frame through — fence and rally set their own frames and stay untouched. If you prefer to keep it in Rust, gate it on `mission_type == MAV_MISSION_TYPE_MISSION` and take the list verbatim from ArduPilot's `stored_in_location()` with a comment saying so. Either way, please add a fence + rally write round-trip to the test list.

Small things: `de.json` for the four new keys — here is the wording, matching the existing `comRcInModeHint`:

```json
"comRcInModeBlocked": "COM_RC_IN_MODE = {value} auf diesem Fahrzeug: PX4 ignoriert Joystick-Eingaben. Auf 2 (RC und Joystick) setzen, um von hier zu fliegen.",
"comRcInModeOk": "COM_RC_IN_MODE = {value}: Joystick-Eingabe erlaubt. Die Sticks wirken in manuellen Modi (Position / Altitude / Stabilized), nicht in Hold, Takeoff oder Mission.",
"comRcInModeRead": "Neu lesen",
"comRcInModeFix": "Joystick erlauben (auf 2 setzen)"
```

I'll run the ArduPilot SITL round trips (mission with DO items, fence, rally, take-off / land) on my side once the frame change is in.

dldnjsxo95:
Thanks — the fence / rally breakage and the wrong ids were both real, and the altitude one would have been nasty. Two commits pushed:

**`fix(mission): decide the action-item frame in the planner, not in the upload path`** — your first suggestion. `wp_to_item()` passes the frame through unchanged again (`cmd_has_location()` is gone), so fence and rally items keep the GLOBAL / RELATIVE frames their editors set. The decision now lives in `missionArdupilot.ts` next to the catalog: `framesForUpload()` sets `MAV_FRAME_MISSION` on every command the catalog knows that is *not* in ArduPilot's `stored_in_location()` set — taken verbatim with a comment (16-19, 21, 22, 30, 31, 82, 84, 85, 92, 94, 179, 188, 189, 191, 195, 201; RTL 20 kept in the pass-through set so it goes out as before). Commands the catalog does not know keep the frame they came with, for exactly the reason you gave — a wrong guess turns a relative altitude into an absolute one on ArduPilot. It runs only in `ArduMissionPanel`'s upload. The PX4 NaN-yaw handling is untouched.

**`i18n(de): COM_RC_IN_MODE banner strings`** — your wording, verbatim.

Verified against PX4 SITL: mission upload through `framesForUpload` sends `[3, 3, 2, 3, 3]` for TAKEOFF / WP / DO_CHANGE_SPEED (file frame 0) / WP / LAND, accepted, 5/5 back with the DO item in frame 2. **Fence write round trip**: one inclusion polygon (4 vertices) + `GF_ACTION = 1` written, read back identical, cleared, `GF_ACTION` restored. **Rally write round trip**: one point written, read back identical, cleared. Nothing on the PX4 console. Both round trips are in the PR's test list now; the ArduPilot SITL side (commits 1, 6, 8) is yours as agreed — the fence/rally write is the case to hit there.

b14ckyy (changes requested):
The frame move is exactly right — checked the location set against current ArduPilot master `stored_in_location()` and against our catalog: every location command the catalog knows is in the set, unknown ones keep their frame, fence and rally are back on their own frames. `de.json` matches, `cargo test` and `svelte-check` pass on the head.

Two things from commits 4 and 7 that I only caught on a second pass — both small, but the first one is user-visible, so one more round please:

1. `fence_write_all` / `rally_write_all` call `control::set_param` with `?` **after** the geometry upload succeeded. On PX4 that call now does a typed PARAM read first and returns `Err` on a timeout, so a slow parameter read leaves the editor "dirty" although the fence is already on the vehicle. Collect the parameter errors and report them as a warning instead of failing the save; the "fire-and-forget" doc comment on `set_param` is stale as well.
2. `RcControlPanel` reads `COM_RC_IN_MODE` in a `$effect` the moment a PX4 vehicle is connected — in parallel with the sequential fence/rally read on connect. The handler has a single `param_fwd` slot and `params_rt` unregisters unconditionally, so two concurrent readers can starve each other for 3 s. Put the read into the on-connect sequence in `connectionController.ts` instead of the effect.

Two notes, no action needed: `AUTOPILOT_COMPONENT` is hard-coded to 1 although the handshake stores `fc_compid`; and `cmd_yaw_in_param4` includes 18/19 (param4 is xtrack) and 82 (param4 unused) — `16 | 17 | 21 | 22 | 84 | 85` is the yaw set.

ArduPilot SITL round trips (mission with DO items, fence, rally, take-off / land) run on my side after that.

PR #196, opened 2026-09-15
fix(hid): Windows — XInput fallback for an Xbox-class pad that Windows.Gaming.Input lists but never reads (+ [hid] diagnostics)

dldnjsxo95:
## Description

Windows game-controller input: when Windows.Gaming.Input (WGI) lists an Xbox-class pad but never delivers a reading, read it through the classic XInput API instead. Plus `[hid]` diagnostics in the backend log so the next report of this kind can be diagnosed from the log alone. Split out of the multi-vehicle branch (#136) as requested.

### The failure, as requested

| | |
|---|---|
| Pad | Third-party Xbox-class (GIP) controller, wired USB. USB `VID_11C1&PID_2001` (rev 0104), bus-reported name "Controller", WGI display name "Xbox One Game Controller". Compatible id `USB\MS_COMP_XGIP10`. |
| Driver stack | `dc1-controller` **10.0.26100.9444** (Microsoft, 2026-09-03) for the GIP device; it exposes the standard XInput HID `VID_045E&PID_02FF&IG_00` via `xinputhid` 10.0.26100.9278. |
| Windows | Windows 11 **25H2**, build 26200.9445. |
| What WGI does | `RawGameController.RawGameControllers` lists the pad (6 axes, 14 buttons, 0 switches). `Gamepad.Gamepads` is **empty**. `GetCurrentReading` returns **timestamp 0 on every call** — 384 samples over 12 s from a foreground window in a standalone probe, and Kite's own thread sees the same (see the `[hid]` lines below). |
| What XInput does | `XInputGetState(0)` returns `ERROR_SUCCESS` with a packet number and live axes for the very same pad. |
| Other machines / builds | Only this machine was available — the same pad has **not** been tried on another Windows build. Your Xbox pads working through WGI is consistent with what we see: the fallback never engages for a pad that `Gamepad.Gamepads` lists. |

### What the fallback does

- Only an entry with the Xbox raw layout (6 / 14 / 0) that is **still silent 1.5 s after it appeared AND absent from `Gamepad.Gamepads`** is switched to XInput. The second condition is the discriminator: a working stack lists every Xbox-class pad in `Gamepad.Gamepads` within a few hundred milliseconds, so a resting pad there (WGI reports timestamp 0 until its first input) stays on WGI. A pad with any other layout (HOTAS, RC transmitter, DualShock) never enters this path.
- The WGI object is kept for identity (name, stable id); only the live state comes from `XInputGetState`. The k-th Xbox-class entry maps to the k-th connected XInput user — exact with one pad.
- Axes/buttons are emitted in WGI's raw order for Xbox pads (LX LY RX RY LT RT; Menu View A B X Y D-pad ↑↓←→ LB RB LS RS), triggers remapped 0..1 → −1..1 like WGI does, so a mapping learned on one source keeps working if the other takes over.
- If the XInput slot vanishes (unplug), the entry goes back to WGI with the grace period re-armed.
- Cargo: enables the `Win32_UI_Input_XboxController` feature of the `windows` crate (already a dependency).

### `[hid]` diagnostics (first commit, all platforms)

Thread start/stop, every listed device (id, name, axes/buttons/hats, uuid), which device is streamed, the first reading from it, a 1 Hz debug trace of the axis vector while it changes, and a one-time warning after 5 s of silence with the foreground-window hint. Log lines only.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Build / CI

## Checklist

- [x] `npm run check` passes — no frontend changes
- [x] `cargo check` (in `src-tauri`) passes — clippy clean in the touched files, `cargo test` green
- [x] I have tested the changes locally (dev, the machine above). Backend log after starting the RC input thread with the pad at rest:
  ```
  [hid] input thread started
  [hid] device id=0 'Xbox One Game Controller' axes=6 buttons=14 hats=0 ({wgi/nrid/…})
  [hid] streaming device id=0
  [hid] 'Xbox One Game Controller' delivers no Windows.Gaming.Input readings — reading it through XInput (user 0)   ← 1.5 s later
  [hid] first reading from device id=0 (6 axes, 14 buttons, 0 hats)
  [hid] axes%=[0, 0, 0, 0, -100, -100] buttons=[]
  ```
  Standalone probes on the same machine: WGI `RawGameControllers` = 1 (`vid=0x11C1 pid=0x2001 axes=6 btn=14 sw=0`), `Gamepads` = 0, 384 `GetCurrentReading` calls over 12 s all timestamp 0; `XInputGetState(0)` = connected with a packet number. Stick movement reaching the RC mapping through the XInput path was verified when the fallback was written (2026-09-11); today's run is the resting vector.
- [ ] Documentation (ROADMAP / CHANGELOG) updated if needed — changelog line suggested below
- [x] No unrelated changes

## Notes for reviewer

- Two commits: `feat(hid)` diagnostics (all platforms, log lines only) and `fix(hid)` the Windows fallback.
- The thing to confirm on your Xbox pads: `Gamepad.Gamepads` lists them (it does on every stack we know of), the `[hid] first reading from device` line arrives after the first stick movement, and **no** `reading it through XInput` line ever appears. If it does, `Gamepad.Gamepads` was empty on your machine too — tell me and we will look at why.
- Not tested: two Xbox-class pads at once (the k-th-entry ↔ k-th-user mapping), wireless / adapter variants.
- Suggested changelog line: "Windows: an Xbox-class gamepad that Windows.Gaming.Input lists but never reads is read through XInput; the backend log now shows what the game-controller thread sees."


b14ckyy (approved):
Read both commits against the current backend — this is clean, and I can't see a way for it to touch a pad that works through WGI: the fallback only lives in the existing `timestamp == 0` dead end and needs Xbox layout **and** 1.5 s of silence **and** an empty `Gamepad.Gamepads` at the same time, the rescan keeps `source` / `since` across its 500 ms cycle (entries are reused by NonRoamableId), and a lost XInput slot falls back to WGI with the grace period re-armed. The diagnostics follow the log-level convention exactly (device list only on change, one `warn` after 5 s of silence at the default level, the axis trace at `debug`). Approving; the hands-on test with our own Xbox pads happens after the merge.

Two notes, neither blocks:

1. **One WinRT call per tick while a working pad rests.** After the 1.5 s, `snapshot()` calls `Gamepad::Gamepads()` on every 20 ms poll for an Xbox-layout entry that hasn't produced a reading yet — on a working stack that is every tick until the first stick movement. The `&&` order is right (XInput is never reached there), so it is only cost, not behaviour. Either cache the `Gamepads` result in `rescan()` (500 ms is plenty for this decision) or throttle the check to a couple of Hz — or tell us why per-tick is needed and we leave it.
2. **Axis sign parity between the two sources is unverified.** WGI raw usually reports "stick up" as 0.0 (→ −1 after the remap), XInput as +32767 (→ +1). If so, LY/RY flip between WGI and XInput, which only matters for a mapping learned on WGI before the driver stack broke. You can't compare that on your machine (WGI never reads there), so no action — we'll check it on our side after the merge if it comes up, and the changelog line can say "re-learn the mapping if the sticks feel inverted".

dldnjsxo95:
Thanks for the careful read. Note 1 is in as a separate commit (`perf(hid): sample Gamepad.Gamepads once per rescan instead of every poll tick`): the flag is sampled in `rescan()` on the existing 500 ms cadence and `snapshot()` reads the cached value, so a resting Xbox-layout entry costs no WinRT call per tick any more. Behaviour unchanged; `cargo check` / clippy clean.

Note 2 — agreed, I cannot compare the two sources here. Happy with the changelog wording "re-learn the mapping if the sticks feel inverted"; if your pads show the LY/RY flip I can add a per-source sign table in a follow-up.

b14ckyy:
Re-read the perf commit: the flag is sampled in `rescan()` on the 500 ms cadence before every `snapshot()`, so it is never stale for the decision and the initial `false` is covered by the first poll. Behaviour identical, cost gone. CI is green on all four targets and the test suite passes locally on Windows. Merging — thanks for the quick turnaround. The axis-sign check happens on our pads after the merge; if they flip, we'll take you up on the sign table.

